PR #9, opened 2026-01-04
fix: Add unregister() method to BundleRegistry for bundle removal

robotdad:
## Summary

- Implements `BundleRegistry.unregister(name)` method to enable proper removal of bundles from the registry
- Fixes bug where `amplifier bundle remove` doesn't fully remove bundles from the in-memory registry
- Method removes bundle from `_bundles` dict and persists changes to disk via `_save_index()`

## Companion PR

⚠️ **This PR pairs with a companion PR in amplifier-app-cli** that uses this new `unregister()` method to complete the fix for `amplifier bundle remove`.

## Test plan

- [x] Added `unregister()` method after `list_registered()` in registry.py
- [x] Method removes bundle from in-memory registry
- [x] Method persists changes via `_save_index()`
- [x] Returns True if removed, False if not found
- [ ] Integration testing will be performed with the companion app-cli PR

### DIFF
--- a/amplifier_foundation/registry.py
+++ b/amplifier_foundation/registry.py
@@ -192,8 +192,9 @@ class BundleRegistry:
     def unregister(self, name: str) -> bool:
         """Remove a bundle from the registry.
 
-        Removes the bundle from the in-memory registry and persists the change
-        to disk. This does not delete cached files.
+        Removes the bundle from the in-memory registry.
+        Does NOT persist automatically - call save() to persist.
+        This does not delete cached files.
 
         Args:
             name: Bundle name to remove.
@@ -204,8 +205,23 @@ class BundleRegistry:
         if name not in self._registry:
             return False
 
+        state = self._registry[name]
+
+        # Clean up included_by refs in bundles we include
+        if state.includes:
+            for child_name in state.includes:
+                child = self._registry.get(child_name)
+                if child and child.included_by:
+                    child.included_by = [n for n in child.included_by if n != name]
+
+        # Clean up includes refs in bundles that include us
+        if state.included_by:
+            for parent_name in state.included_by:
+                parent = self._registry.get(parent_name)
+                if parent and parent.includes:
+                    parent.includes = [n for n in parent.includes if n != name]
+
         del self._registry[name]
-        self.save()
         logger.debug(f"Unregistered bundle: {name}")
         return True
 
--- a/amplifier_foundation/registry.py
+++ b/amplifier_foundation/registry.py
@@ -189,6 +189,26 @@ class BundleRegistry:
         """
         return sorted(self._registry.keys())
 
+    def unregister(self, name: str) -> bool:
+        """Remove a bundle from the registry.
+
+        Removes the bundle from the in-memory registry and persists the change
+        to disk. This does not delete cached files.
+
+        Args:
+            name: Bundle name to remove.
+
+        Returns:
+            True if bundle was found and removed, False if not found.
+        """
+        if name not in self._registry:
+            return False
+
+        del self._registry[name]
+        self.save()
+        logger.debug(f"Unregistered bundle: {name}")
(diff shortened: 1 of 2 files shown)


bkrabach (changes requested):
## Review: BundleRegistry.unregister()

Thanks for tackling this bug fix! The approach is correct, but there are a few issues to address before merging.

### 1. Missing Tests (Required)

The PR adds functionality but no tests. `test_registry.py` has comprehensive coverage for other methods - please add tests following the same patterns:

```python
class TestUnregister:
    def test_unregister_existing_bundle_returns_true(self): ...
    def test_unregister_nonexistent_bundle_returns_false(self): ...
    def test_unregister_removes_from_list_registered(self): ...
```

### 2. Remove Auto-Save (Required)

`unregister()` currently calls `self.save()`, but `register()` does not - by design. The docstring for `register()` explicitly states:

> "Does NOT persist automatically - call save() to persist."

The `registry.json` is a **cache of loaded bundle state**, not the source of truth (that's `user_registry` in the CLI layer). Cache semantics should be consistent: the caller controls when to persist.

Please remove the `self.save()` call:

```python
def unregister(self, name: str) -> bool:
    if name not in self._registry:
        return False
    
    del self._registry[name]
    logger.debug(f"Unregistered bundle: {name}")
    return True
    # Caller can call save() if persistence is needed
```

The CLI's `bundle_remove()` can call `registry.save()` after `registry.unregister()` if persistence is desired.

### 3. Relationship Cleanup (Recommended)

`BundleState` tracks `includes` and `included_by` relationships. When unregistering, these references become stale in other bundles. Consider:

```python
def unregister(self, name: str) -> bool:
    if name not in self._registry:
        return False
    
    state = self._registry[name]
    
    # Clean up included_by refs in bundles we include
    if state.includes:
        for child_name in state.includes:
            child = self._registry.get(child_name)
            if child and child.included_by:
                child.included_by = [n for n in child.included_by if n != name]
    
    # Clean up includes refs in bundles that include us
    if state.included_by:
        for parent_name in state.included_by:
            parent = self._registry.get(parent_name)
            if parent and parent.includes:
                parent.includes = [n for n in parent.includes if n != name]
    
    del self._registry[name]
    logger.debug(f"Unregistered bundle: {name}")
    return True
```

### Summary

| Item | Priority |
|------|----------|
| Add unit tests | Required |
| Remove `self.save()` call | Required |
| Add relationship cleanup | Recommended |

The companion CLI PR (#34) will need a minor update to call `registry.save()` after `registry.unregister()` if persistence is desired.

robotdad:
## Changes Made to Address Review Feedback

Thank you for the thorough review! I've addressed all the feedback:

### 1. ✅ Removed Auto-Save (Required)

Removed the `self.save()` call from `unregister()` to match `register()` semantics. The docstring now explicitly states:

> "Does NOT persist automatically - call save() to persist."

The caller now controls when to persist, maintaining consistent cache behavior across the registry API.

### 2. ✅ Added Relationship Cleanup (Recommended)

Implemented cleanup logic that removes stale references when unregistering a bundle:

- Cleans up `included_by` references in bundles that this bundle includes
- Cleans up `includes` references in bundles that include this bundle

This prevents orphaned relationship references in the registry.

### 3. ✅ Added Unit Tests (Required)

Added `TestUnregister` class with 7 comprehensive tests following existing patterns:

- `test_unregister_existing_bundle_returns_true`
- `test_unregister_nonexistent_bundle_returns_false`
- `test_unregister_removes_from_list_registered`
- `test_unregister_does_not_auto_persist`
- `test_unregister_cleans_up_includes_relationships`
- `test_unregister_cleans_up_included_by_relationships`
- `test_unregister_handles_partial_relationships`

**All tests pass** ✅

---

The companion CLI PR will be updated to call `registry.save()` after `registry.unregister()` as needed.

Ready for re-review!

PR #31, opened 2026-01-21
fix: extract agent markdown body as system instruction

samueljklee:
## Problem

When loading agent `.md` files via `_load_agent_file_metadata()`, the markdown body after frontmatter was being discarded (assigned to `_body` and ignored). This body contains the agent's behavioral instructions that should be injected as the system prompt when the agent is spawned via the task tool.

**Fixes**: microsoft/amplifier#174

## Root Cause

Two issues combined to cause this bug:

1. **`_load_agent_file_metadata()` discarded the body**: The markdown content after frontmatter was parsed but assigned to `_body` (unused variable) and never stored.

2. **`load_agent_metadata()` was never called**: The method existed but `Bundle.prepare()` never invoked it before generating the mount plan.

## Impact

Agents like `smoke-tester` that rely on their markdown instructions to guide behavior (e.g., "run this exact recipe path") would receive no system prompt and fall back to generic exploration behavior, causing unpredictable results.

## Solution

1. Modified `_load_agent_file_metadata()` to capture the body and store it as `system.instruction`
2. Added call to `self.load_agent_metadata()` in `Bundle.prepare()` before `to_mount_plan()`

## Changes

- `amplifier_foundation/bundle.py`:
  - `prepare()`: Added call to `self.load_agent_metadata()` before `to_mount_plan()`
  - `_load_agent_file_metadata()`: Changed `_body` to `body`, added logic to store as `result["system"] = {"instruction": body.strip()}`

## Tests

Added 3 tests in `TestAgentMetadataLoading`:
- `test_load_agent_metadata_extracts_system_instruction` - verifies body becomes `system.instruction`
- `test_load_agent_metadata_preserves_existing_system` - verifies explicit inline instructions aren't overwritten
- `test_load_agent_metadata_empty_body_no_system` - verifies empty body doesn't create spurious instruction

All 21 tests pass.

---

### DIFF
--- a/amplifier_foundation/bundle.py
+++ b/amplifier_foundation/bundle.py
@@ -252,6 +252,10 @@ class Bundle:
         """
         from amplifier_foundation.modules.activator import ModuleActivator
 
+        # Load full agent metadata (including system instructions from .md files)
+        # Must happen before to_mount_plan() so agent configs include system.instruction
+        self.load_agent_metadata()
+
         # Get mount plan
         mount_plan = self.to_mount_plan()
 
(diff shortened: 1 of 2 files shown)


samueljklee:
Fixed in #32 .

PR #63, opened 2026-02-04
feat: [foundation] - Integrated Session Spawning & Event-Driven Orchestration

payneio:
## Summary
Part of the Integrated Session Spawning & Event-Driven Orchestration implementation.

## Changes
- Implement `spawn_bundle()` unified session spawning primitive
- Enhance `spawn_bundle()` with `metadata_extra` and `system_instruction` support
- Add `EventRouter` for cross-session event communication
- Add trigger infrastructure and `BackgroundSessionManager`

## Dependencies
- Depends on microsoft/amplifier-core PR for `SESSION_COMPLETED`/`SESSION_ERROR` events

## Testing
- All unit tests pass
- Validation recipes pass

### DIFF
--- a/amplifier_foundation/__init__.py
+++ b/amplifier_foundation/__init__.py
@@ -81,6 +81,11 @@ from amplifier_foundation.spawn_utils import apply_provider_preferences
 from amplifier_foundation.spawn_utils import apply_provider_preferences_with_resolution
 from amplifier_foundation.spawn_utils import is_glob_pattern
 from amplifier_foundation.spawn_utils import resolve_model_pattern
+
+# Spawn primitive
+from amplifier_foundation.spawn import SessionStorage
+from amplifier_foundation.spawn import SpawnResult
+from amplifier_foundation.spawn import spawn_bundle
 from amplifier_foundation.sources.protocol import SourceHandlerProtocol
 from amplifier_foundation.sources.protocol import SourceHandlerWithStatusProtocol
 from amplifier_foundation.sources.protocol import SourceResolverProtocol
@@ -179,4 +184,8 @@ __all__ = [
     "apply_provider_preferences_with_resolution",
     "is_glob_pattern",
     "resolve_model_pattern",
+    # Spawn primitive
+    "SessionStorage",
+    "SpawnResult",
+    "spawn_bundle",
 ]
--- a/amplifier_foundation/__init__.py
+++ b/amplifier_foundation/__init__.py
@@ -86,6 +86,11 @@ from amplifier_foundation.spawn_utils import resolve_model_pattern
 from amplifier_foundation.spawn import SessionStorage
 from amplifier_foundation.spawn import SpawnResult
 from amplifier_foundation.spawn import spawn_bundle
+
+# Event routing
+from amplifier_foundation.events import EventRouter
+from amplifier_foundation.events import SessionEmitter
+from amplifier_foundation.events import SessionEvent
 from amplifier_foundation.sources.protocol import SourceHandlerProtocol
 from amplifier_foundation.sources.protocol import SourceHandlerWithStatusProtocol
 from amplifier_foundation.sources.protocol import SourceResolverProtocol
@@ -188,4 +193,8 @@ __all__ = [
     "SessionStorage",
     "SpawnResult",
     "spawn_bundle",
+    # Event routing
+    "EventRouter",
+    "SessionEvent",
+    "SessionEmitter",
 ]
--- a/amplifier_foundation/__init__.py
+++ b/amplifier_foundation/__init__.py
@@ -91,6 +91,19 @@ from amplifier_foundation.spawn import spawn_bundle
 from amplifier_foundation.events import EventRouter
 from amplifier_foundation.events import SessionEmitter
 from amplifier_foundation.events import SessionEvent
+
+# Triggers
+from amplifier_foundation.triggers import ManualTrigger
+from amplifier_foundation.triggers import SessionEventTrigger
+from amplifier_foundation.triggers import TimerTrigger
+from amplifier_foundation.triggers import TriggerEvent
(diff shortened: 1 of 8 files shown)


bkrabach:
Thanks for this work, @payneio — this surfaces real architectural gaps, particularly around cross-session communication and spawn infrastructure. After a thorough investigation, we think the right path is to build this as a standalone bundle rather than adding it to the foundation library. Here's our reasoning:

**The public APIs already support what you need**

The core and foundation primitives your `spawn_bundle()` uses are all publicly accessible today:
- `load_bundle()` + `bundle.prepare()` for bundle resolution
- `AmplifierSession(config, session_id, parent_id)` for session creation
- `coordinator.mount()`, `coordinator.register_capability()` for infrastructure wiring
- `coordinator.get("context")` + `add_message()` for context inheritance
- `coordinator.cancellation` for cancellation propagation
- `session.initialize()`, `session.execute()`, `session.cleanup()` for the lifecycle

You can build your own spawn function on top of these in your own bundle — which is effectively what `spawn_bundle()` already does.

**Why not in foundation**

Foundation's role is bundle primitives and shared utilities. This PR adds 1,960 lines and 24 new exports without proven consumers yet. The kernel philosophy's two-implementation rule applies here: prove ideas as modules first, promote to foundation after multiple independent consumers converge on the need.

Applying the mechanism/policy lens:
- `EventRouter` is a legitimate mechanism, but with zero consumers today it's premature for foundation
- `TimerTrigger`, `SessionEventTrigger`, `ManualTrigger` are specific strategies — policy that belongs at the edges
- `BackgroundSessionManager` is lifecycle policy — a strong candidate for a behavior bundle

**What we'd suggest**

Package all of this as `amplifier-bundle-background-sessions` (or similar). This is exactly the pattern foundation follows — `amplifier-bundle-recipes`, `amplifier-bundle-shadow`, etc. are all capability bundles that compose on top of foundation without modifying it. If the pattern proves out and other modules converge on needing `EventRouter`, we can promote it to foundation later.

**On the spawn gaps you found**

Your `spawn_bundle()` highlighted real missing pieces in `PreparedBundle.spawn()` — cancellation propagation, mention resolver/deduplicator capability registration, bundle_package_paths, timeout enforcement. We'll address these in a separate foundation PR as surgical fixes to the existing spawn mechanism. Those fixes will also benefit your bundle.

**On the core dependency**

We closed core PR #9 — the existing `SESSION_END = "session:end"` constant already covers the need. The kernel shouldn't emit it (it doesn't know when a session is truly done — only the caller does), but your bundle can emit it through the existing hooks system when it knows a spawned session has completed.

Closing this PR, but the work isn't wasted — it maps directly to a standalone bundle. Let us know if you want to discuss the bundle structure.

payneio:
## Response to Feedback

Thank you for the detailed architectural review. We agree with several points but would like to discuss the `spawn_bundle()` placement.

### What We Accept ✅

| Component | Your Feedback | Our Response |
|-----------|---------------|--------------|
| `TimerTrigger`, `SessionEventTrigger`, `ManualTrigger` | Policy, belongs at edges | Agree - will move to bundle |
| `BackgroundSessionManager` | Lifecycle policy | Agree - will move to bundle |
| `EventRouter` | Premature for foundation | Agree for now - will move to bundle, can promote later if needed |

### Where We Would Like to Discuss Further

**The "zero consumers" characterization does not match our understanding.**

We have **three identified consumers** for `spawn_bundle()`:

| Consumer | Use Case | Status |
|----------|----------|--------|
| **tool-delegate** | Agent delegation via `session.spawn` capability | Active - CLI `spawn_sub_session()` already calls `spawn_bundle()` internally |
| **foreman-bundle** | Worker pools, background sessions | Target consumer - original driver for this work |
| **observers-bundle** | Trigger-based session spawning | Target consumer - original driver for this work |

This work started because we identified **duplication across these three consumers**. Each needs:
- Inheritance logic (providers, tools, hooks with filtering)
- Context depth/scope filtering
- W3C trace-based session IDs
- Cancellation propagation
- @mention resolver inheritance
- Timeout enforcement

This appears to meet the two-implementation rule — multiple independent consumers converging on the same need.

### The "Public APIs Support What You Need" Point

We agree the primitives exist. However, building spawning correctly requires ~100+ lines of careful orchestration:

```python
bundle = await load_bundle(...)
prepared = await bundle.prepare()
session = AmplifierSession(config, session_id, parent_id)
await session.coordinator.mount(...)
await session.coordinator.register_capability("mention_resolver", ...)
await session.coordinator.register_capability("session.working_dir", ...)
# ... handle inheritance, cancellation linking, context injection, etc.
```

Without a blessed primitive, each consumer reimplements this differently, leading to subtle inconsistencies (which is exactly what we found when analyzing delegate vs foreman vs observers).

### Proposed Compromise

1. **Keep `spawn_bundle()` in foundation** — It is the mechanism all three consumers need
2. **Move to `amplifier-bundle-orchestration`:**
   - `EventRouter`
   - `TimerTrigger`, `SessionEventTrigger`, `ManualTrigger`
   - `BackgroundSessionManager`
3. **Address spawn gaps you identified** — We agree these need fixing in the spawn primitive (whichever one is blessed):
   - System prompt factory (not raw `system_instruction`)
   - Cancellation propagation
   - @mention resolver/deduplicator inheritance
   - Timeout enforcement

### Alternative Question

If `spawn_bundle()` must move out of foundation, we would like clarity on:

1. Should the "surgical fixes" to `PreparedBundle.spawn()` include all features from `spawn_bundle()`?
2. Is `PreparedBundle.spawn()` the blessed primitive for delegate/foreman/observers to use?
3. How should consumers handle the differences (e.g., `PreparedBundle.spawn()` uses bundle composition while `spawn_bundle()` uses runtime inheritance)?

Happy to discuss further or jump on a call to align on the right path forward.

PR #68, opened 2026-02-08
feat: add strict mode for bundle validation

ramparte:
## Summary

Adds an optional `strict` mode to bundle validation that promotes include failures from warnings to errors. When strict mode is enabled, missing or unresolvable includes cause `BundleDependencyError` instead of being silently skipped.

## Motivation

The amplifier-distro project needs bundle validation to fail loudly when includes are broken, rather than silently degrading. This is the foundation-level enablement for `strict: true` in distro.yaml.

## Changes

**`registry.py`**
- `BundleRegistry.__init__()` accepts `strict: bool = False`
- `_compose_includes()` raises `BundleDependencyError` in strict mode instead of warning+skip
- `load_bundle()` threads `strict` parameter through to registry

**`validator.py`**
- `BundleValidator.__init__()` accepts `strict: bool = False`
- Missing context paths become errors (not warnings) when strict

**Tests**
- 10 new tests across `test_registry.py` and `test_validator.py`
- All 49 tests pass

## Usage

```python
# Default behavior unchanged (non-strict)
bundle = await load_bundle("my-bundle")

# Strict mode - broken includes raise exceptions
bundle = await load_bundle("my-bundle", strict=True)

# Validator strict mode
validator = BundleValidator(strict=True)
result = validator.validate(bundle)  # missing context = error, not warning
```

## Test plan

- [x] 10 new tests for strict mode behavior in registry and validator
- [x] All 49 existing tests continue to pass (no regression)
- [x] Default (non-strict) behavior unchanged

### DIFF
--- a/amplifier_foundation/registry.py
+++ b/amplifier_foundation/registry.py
@@ -24,13 +24,14 @@ import logging
 from dataclasses import dataclass
 from datetime import datetime
 from pathlib import Path
-from typing import TYPE_CHECKING
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
 from amplifier_foundation.bundle import Bundle
-from amplifier_foundation.exceptions import BundleDependencyError
-from amplifier_foundation.exceptions import BundleLoadError
-from amplifier_foundation.exceptions import BundleNotFoundError
+from amplifier_foundation.exceptions import (
+    BundleDependencyError,
+    BundleLoadError,
+    BundleNotFoundError,
+)
 from amplifier_foundation.io.frontmatter import parse_frontmatter
 from amplifier_foundation.io.yaml import read_yaml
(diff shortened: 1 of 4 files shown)


bkrabach:
Thank you for this work, @ramparte — the strict mode design is clean and well-scoped. We've built on it in PR #69, which includes your original implementation plus two additional improvements:

1. **Phase 2 strict coverage** — include load failures during `asyncio.gather` (e.g., malformed YAML in a resolved include) now also respect strict mode
2. **`load_bundle()` API safety** — raises `ValueError` if both `strict=True` and `registry` are provided, preventing silent parameter ignore

Your commit is preserved as the first commit in #69. Closing this in favor of that PR.

PR #62, opened 2026-02-03
feat: add OpenAI-optimized foundation bundle and context

DavidKoleczek:
## Summary

- Add `context/shared/common-openai-base.md`: OpenAI-specific system instructions with editing constraints, task management guidelines, frontend coding rules, pre-commit validation, and presentation formatting
- Add `experiments/exp-foundation-openai.md`: Experimental foundation bundle configuration optimized for OpenAI models (removes agents behavior, references the new OpenAI context file)

## Details

This PR introduces an experimental bundle configuration tailored for OpenAI models. The key changes include:

1. **OpenAI-optimized context file** (`common-openai-base.md`):
   - Editing constraints (ASCII preference, code comments guidelines)
   - Git worktree handling instructions
   - Task management behaviors
   - Frontend/web development guidelines
   - Pre-commit validation rules
   - Response presentation formatting

2. **Experimental bundle** (`exp-foundation-openai.md`):
   - Based on the standard foundation bundle
   - Removes the agents behavior for a simpler configuration
   - References the new OpenAI-specific context file

## Test plan

- [ ] Verify bundle loads correctly with `amplifier chat --bundle experiments/exp-foundation-openai.md`
- [ ] Confirm context file is properly included in the system prompt
- [ ] Test with OpenAI models to validate instructions are followed

🤖 Generated with [Amplifier](https://github.com/microsoft/amplifier)

PR #83, opened 2026-02-15
fix: update stale nested bundle URIs when reloaded from different source

colombod:
## Summary

Fixes the persistent-state half of [microsoft-amplifier/amplifier-support#62](https://github.com/microsoft-amplifier/amplifier-support/issues/62).

When a parent bundle (e.g., `foundation`) is overridden to a local path via `settings.local.yaml`, nested bundles previously persisted with `git+` URIs kept their stale URIs indefinitely. This PR ensures registry entries are refreshed when a bundle is reloaded from a different source.

## Root Cause

`BundleRegistry._load_single()` has an update block (lines 506–510) that refreshes `version`, `loaded_at`, and `local_path` for known bundles — but it **did not update `uri`**. Combined with persisted state loaded from `registry.json` on init, this meant:

1. First session (no overrides): nested bundles get `git+` URIs → persisted to `registry.json`
2. Second session (with local overrides): `_load_single()` reloads the nested bundle from a `file://` URI, but the existing entry's `uri` field is never updated
3. `registry.json` retains the stale `git+` URI across all subsequent sessions

## Change

**+5 lines in `amplifier_foundation/registry.py`** — surgical addition to the existing update block:

```python
if update_name:
    state = self._registry[update_name]
    if state.uri != uri:                          # ← NEW
        logger.debug(                             # ← NEW
            f"Updating URI for '{update_name}': {state.uri} → {uri}"
        )                                         # ← NEW
        state.uri = uri                           # ← NEW
    state.version = bundle.version
    state.loaded_at = datetime.now()
    state.local_path = str(local_path)
```

The URI is only updated when it actually differs, with a debug log for traceability. For same-URI reloads this is a no-op.

### Diff stats
- `amplifier_foundation/registry.py`: **+5 lines** (production fix)
- `tests/test_registry.py`: **+164 lines** (3 regression tests)

### Test results

| Suite | Before | After | Delta |
|---|---|---|---|
| `amplifier-foundation` full suite | 298 passed ✅ | 301 passed ✅ | +3 new, 0 regressions |

### New regression tests (`TestNestedBundleURIUpdate`)

| Test | What it verifies |
|---|---|
| `test_persisted_nested_bundle_uri_updates_on_reload` | Pre-populated `git+` URI in `registry.json` gets updated to `file://` when the same bundle is loaded from a local path |
| `test_persisted_uri_update_survives_save_reload` | Updated URI persists correctly across `save()` → new `BundleRegistry` cycle |
| `test_same_uri_reload_does_not_change_entry` | Reloading from the same URI is a no-op (guards against unnecessary mutations) |

## Companion PR

This is the first of two PRs for issue #62. The companion PR in `amplifier-app-cli` changes `get_bundle_categories()` to read from the in-memory registry instead of stale `registry.json` on disk. **This foundation PR should merge first** (it fixes the persistent state that the CLI reads).

## How to verify end-to-end

After both PRs merge:
1. Add `bundle.added.foundation: ../../amplifier-foundation` to `settings.local.yaml`
2. Run any Amplifier session (to trigger bundle loading with the override)
3. Run `amplifier bundle list --all`
4. Confirm nested bundles show `file://` paths, not `git+` GitHub URLs

### DIFF
--- a/amplifier_foundation/registry.py
+++ b/amplifier_foundation/registry.py
@@ -505,6 +505,11 @@ class BundleRegistry:
             )
             if update_name:
                 state = self._registry[update_name]
+                if state.uri != uri:
+                    logger.debug(
+                        f"Updating URI for '{update_name}': {state.uri} → {uri}"
+                    )
+                    state.uri = uri
                 state.version = bundle.version
                 state.loaded_at = datetime.now()
                 state.local_path = str(local_path)
--- a/tests/test_registry.py
+++ b/tests/test_registry.py
@@ -794,3 +794,167 @@ class TestLoadBundleConvenience:
 
             with pytest.raises(BundleDependencyError, match="strict mode"):
                 await load_bundle(f"file://{bundle_dir}", strict=True)
+
+
+class TestNestedBundleURIUpdate:
+    """Tests for URI update when a nested bundle is reloaded from a different source.
+
+    Regression tests for https://github.com/microsoft-amplifier/amplifier-support/issues/62
+    When a parent bundle is overridden to a local path, nested bundles previously
+    persisted with git+ URIs must have their registry entries updated.
+    """
+
+    @pytest.mark.asyncio
+    async def test_persisted_nested_bundle_uri_updates_on_reload(self) -> None:
+        """Registry entry URI updates when same bundle loads from a different URI."""
+        import json
+
+        with tempfile.TemporaryDirectory() as tmpdir:
+            base = Path(tmpdir)
+            home = base / "home"
+
+            # Create local bundle files (the "new" local source)
+            local_repo = base / "local-repo"
+            local_repo.mkdir()
+            (local_repo / "bundle.md").write_text(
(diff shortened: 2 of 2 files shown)


sadlilas:
## Review: Approve

Independently verified the root cause by tracing all read sites for `state.uri` in the registry. This is not a display bug -- stale URIs cause functional issues in bundle resolution, include resolution, and update checking.

The fix correctly adds `state.uri = uri` to the existing update block in `_load_single()`. This is consistent with the established pattern -- version, loaded_at, and local_path are already updated there. The URI was the only field missing from the update set.

Verified that `update_name` is always set for persisted bundles. No scenario where a persisted nested bundle would skip this update.

Tests are solid: URI update on reload, persistence across save/load cycles, same-URI idempotency.

Non-blocking note: `amplifier reset --remove cache` does not delete `registry.json` (separate category in reset.py). After cache reset, stale URIs persist until the next bundle load triggers this fix. Worth a follow-up issue.

LGTM. We recommend merging this PR and not merging the companion app-cli PR #92 -- once this fix lands, the first session corrects all persisted URIs and the display self-heals permanently.

PR #95, opened 2026-02-27
fix: git source integrity check now considers #subdirectory= for monorepos

payneio:
## Problem

`_verify_clone_integrity` in `sources/git.py` checks only the **repo root** for `pyproject.toml`, `setup.py`, or `bundle.md`. For monorepo-style bundles using `#subdirectory=` URI fragments (e.g. `git+https://github.com/microsoft/amplifier-distro@main#subdirectory=bundle`), these files live in a subdirectory, not the root.

The clone completes successfully but the integrity check rejects it, causing `BundleNotFoundError` on every load attempt. The cache entry is then deleted, so the failure repeats on every subsequent call.

## Fix

`_verify_clone_integrity` now accepts an optional `subpath` parameter. When provided, it checks both the repo root **and** `cache_path / subpath` for expected files. The two call sites in `resolve()` pass `parsed.subpath` through.

Default behavior (no subpath) is unchanged — this only affects URIs with `#subdirectory=` fragments.

## Testing

- Non-monorepo URIs: behavior identical (subpath defaults to None, only root is checked)
- Monorepo URIs with `#subdirectory=bundle`: now finds `bundle/bundle.md` and passes integrity check
- Corrupted/partial clones: still correctly rejected (missing `.git` dir check unchanged)

### DIFF
--- a/amplifier_foundation/sources/git.py
+++ b/amplifier_foundation/sources/git.py
@@ -104,16 +104,24 @@ class GitSourceHandler:
         with contextlib.suppress(OSError):
             meta_path.write_text(json.dumps(metadata, indent=2, default=str))
 
-    def _verify_clone_integrity(self, cache_path: Path) -> bool:
+    def _verify_clone_integrity(
+        self, cache_path: Path, subpath: str | None = None
+    ) -> bool:
         """Verify that a cloned repository has expected structure.
 
         Checks for indicators that the clone completed successfully and contains
-        a valid Python module. This catches cases where git clone partially
-        succeeds but leaves an incomplete directory (e.g., due to network issues,
-        cloud sync interference, or disk I/O errors).
+        a valid Python module or bundle. This catches cases where git clone
+        partially succeeds but leaves an incomplete directory (e.g., due to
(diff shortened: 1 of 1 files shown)


sadlilas:
Thanks for the PR, @payneio. The code change itself is well-written, but after investigating the full end-to-end picture, we're going to defer this for now.

**What we found:**

The integrity check fix is correct, but it's one piece of what would need to be a broader effort to support rootless monorepo bundles. There's a downstream bug in `_resolve_include_source` (registry.py:861) where namespace includes from subdirectory bundles drop the subdirectory prefix — generating wrong `#subdirectory=` paths. Fixing that requires a data model change to `BundleState`. The documentation (BUNDLE_GUIDE.md) also doesn't describe this layout pattern, so merging the plumbing fix without updating guidance creates an inconsistency.

**For now:**

The recommended approach is to use separate repos (one bundle per repo with `bundle.md` at root) — that's the documented pattern and is fully supported today. We may revisit monorepo support later as a coordinated effort.

Closing for now, but we'll keep this in mind if we decide to pursue monorepo support.

PR #154, opened 2026-04-10
fix(hooks-session-naming): make provider timeout configurable

ramparte:
## Summary

The session naming hook had a **hardcoded 10-second timeout** that caused session initialization to fail when using local Ollama models requiring longer cold-load times from external storage.

## Problem

On the Mac Studio running Ollama with a 234GB model directory on external storage (ai-storage), cold-loading large models (e.g., gemma4:31b at 19GB) takes >10 seconds. This caused every session to fail with:
```
Session naming provider call timed out (10 s) for session X
```

The timeout was hardcoded in the module with no override mechanism.

## Solution

Made the provider timeout configurable:
- Added `provider_timeout: float = 10.0` to SessionNamingConfig dataclass
- Changed hardcoded `timeout=10.0` to `timeout=self.config.provider_timeout` in asyncio.wait_for call
- Updated logger output to show the actual configured timeout
- Maintains backward compatibility (10s default preserved)

## Configuration

Users can now override the timeout in `~/.amplifier/settings.yaml`:

```yaml
overrides:
  hooks-session-naming:
    config:
      provider_timeout: 60  # 60 seconds for local models on external storage
```

## Testing

Verified on Mac Studio:
- Before fix: Session naming timed out on every cold-load
- After fix: Sessions complete successfully with 60s provider timeout
- Model used: Ollama (gemma4:31b from external ai-storage)

## Backward Compatibility

✅ Default timeout remains 10 seconds
✅ No breaking changes to the module interface
✅ Configuration is optional

### DIFF
--- a/modules/hooks-session-naming/amplifier_module_hooks_session_naming/__init__.py
+++ b/modules/hooks-session-naming/amplifier_module_hooks_session_naming/__init__.py
@@ -35,6 +35,7 @@ class SessionNamingConfig:
     max_description_length: int = 200
     max_retries: int = 3
     model_role: str | None = "fast"
+    provider_timeout: float = 10.0
 
 
 INITIAL_NAMING_PROMPT = """You generate names and descriptions for conversation sessions.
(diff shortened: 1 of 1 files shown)


ramparte:
Closing as superseded. `provider_timeout` for hooks-session-naming is already configurable in the shipped module (SessionNamingConfig.provider_timeout, read from config in mount()), and consumers set it via bundle config — e.g. amplifier-next now sets `provider_timeout: 30` with `model_role: null`. No code change needed here. Thanks!

PR #182, opened 2026-04-23
fix(tool-delegate): use register_contributor for observability.events

colombod:
## Summary

Migrates `tool-delegate` from the get-extend-register anti-pattern using `register_capability` to `register_contributor` for announcing `observability.events`.

## Problem

`tool-delegate` currently registers its observable events using a pattern that is invisible to `collect_contributions()`:

```python
# Before — invisible to collect_contributions()
obs_events = coordinator.get_capability("observability.events") or []
obs_events.extend(["delegate:agent_spawned", "delegate:agent_completed", ...])
coordinator.register_capability("observability.events", obs_events)
```

`register_capability` is a singleton-ownership mechanism. Any module that discovers events via `collect_contributions("observability.events")` will never see the delegate events, regardless of mount order.

Authoritative reference: `core:docs/specs/CONTRIBUTION_CHANNELS.md` uses `observability.events` as its **primary worked example** of the contribution channel mechanism — `register_capability` is explicitly the wrong choice for this use case.

## Changes

**`modules/tool-delegate/amplifier_module_tool_delegate/__init__.py`**

```python
# After — discoverable via collect_contributions()
coordinator.register_contributor(
    "observability.events",
    "tool-delegate",
    lambda: [
        "delegate:agent_spawned",
        "delegate:agent_resumed",
        "delegate:agent_completed",
        "delegate:agent_cancelled",
        "delegate:error",
    ],
)
```

Also includes:
- Extraction of `_spawn_new_session()` helper for testability
- `isinstance(orch_section, dict)` guard fixing `AttributeError` when orchestrator config is a bare string (e.g. `"loop-basic"`) — `.get("config")` was called on a string
- Exception handler now logs with `exc_info=True` so downstream "Agent not found" errors surface the real upstream cause

**`modules/tool-delegate/pyproject.toml`**
- `amplifier-core>=0.1.0` → `>=1.0.0` (floor for Rust coordinator with contribution channel support)

**`scripts/lint_observability.py`**
- Updated docstring at line 157 — `tool-delegate` is now the counter-example of the old pattern, not an example of it

## Tests

```
test_delegate_contribution_channel.py    3 passed  — register_contributor migration correctness
test_delegate_spawn_new_session.py       3 passed  — isinstance guard regression tests (new)
test_delegate_event_enrichment.py       14 passed
test_delegate_model_role.py              9 passed
test_delegate_provider_preferences.py    3 passed
test_delegate_surface_status.py          4 passed
─────────────────────────────────────────────────────
Total: 37 passed, 0 regressions
```

## Commit

| Hash | Message |
|------|---------|
| `106d30c` | `fix(tool-delegate): use register_contributor for observability.events` |

### DIFF
--- a/modules/tool-delegate/amplifier_module_tool_delegate/__init__.py
+++ b/modules/tool-delegate/amplifier_module_tool_delegate/__init__.py
@@ -57,18 +57,21 @@ async def mount(coordinator: ModuleCoordinator, config: dict[str, Any] | None = 
     """
     config = config or {}
 
-    # Declare observable lifecycle events for this module
-    obs_events = coordinator.get_capability("observability.events") or []
-    obs_events.extend(
-        [
-            "delegate:agent_spawned",  # When agent sub-session spawned
-            "delegate:agent_resumed",  # When agent sub-session resumed
-            "delegate:agent_completed",  # When agent sub-session completed
-            "delegate:agent_cancelled",  # When parent session is cancelled during delegation
-            "delegate:error",  # When delegation fails
-        ]
+    # Register observable lifecycle events via contribution channel.
+    # Uses register_contributor (not register_capability) so events are
+    # discoverable via collect_contributions("observability.events").
+    # See: core:docs/specs/CONTRIBUTION_CHANNELS.md
+    coordinator.register_contributor(
+        "observability.events",
+        "tool-delegate",
+        lambda: [
+            "delegate:agent_spawned",
+            "delegate:agent_resumed",
+            "delegate:agent_completed",
+            "delegate:agent_cancelled",
+            "delegate:error",
+        ],
     )
-    coordinator.register_capability("observability.events", obs_events)
 
     tool = DelegateTool(coordinator, config)
     await coordinator.mount("tools", tool, name=tool.name)
@@ -849,7 +852,56 @@ Agent usage notes:
                     ProviderPreference.from_dict(p) for p in agent_default_prefs
                 ]
 
-        # Get parent session ID
+        return await self._spawn_new_session(
+            agent_name=agent_name,
+            instruction=instruction,
+            context_depth=context_depth,
+            context_scope=context_scope,
+            context_turns=context_turns,
+            provider_preferences=provider_preferences,
+            hooks=hooks,
+            tool_call_id=tool_call_id,
+            parallel_group_id=parallel_group_id,
+            raw_model_role=raw_model_role,
+            agents=agents,
+        )
+
+    async def _spawn_new_session(
+        self,
+        agent_name: str,
+        instruction: str,
+        context_depth: str,
+        context_scope: str,
+        context_turns: int,
+        provider_preferences: list | None,
+        hooks,
+        *,
+        tool_call_id: str = "",
+        parallel_group_id: str | None = None,
+        raw_model_role: str = "",
+        agents: dict | None = None,
+    ) -> ToolResult:
+        """Spawn a new agent sub-session.
+
+        Core spawn logic extracted for testability. Called by execute() after
+        input validation; may also be called directly in tests.
+
+        Args:
+            agent_name: Agent to delegate to
+            instruction: Task instruction (may include inherited context)
+            context_depth: HOW MUCH context to inherit
+            context_scope: WHICH content to inherit
+            context_turns: Number of recent turns (when context_depth="recent")
+            provider_preferences: Resolved provider preferences list
+            hooks: Hook coordinator for event emission
+            tool_call_id: Orchestrator tool call ID (enriches event payloads)
+            parallel_group_id: Parallel group ID (enriches event payloads)
+            raw_model_role: Raw model role string for routing tracking
+            agents: Agent config dict (defaults to coordinator.config["agents"])
+
+        Returns:
+            ToolResult with spawn outcome
+        """
         parent_session_id = self.coordinator.session_id
 
         # Generate hierarchical sub-session ID (sanitized for filesystem safety)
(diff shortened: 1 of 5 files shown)


bkrabach (changes requested):
# REQUEST CHANGES

The core migration (`register_capability` → `register_contributor` for `observability.events`) is correct and well-tested — 34/34 tool-delegate tests pass, 1248 root tests with zero regressions, and the API usage matches the kernel's canonical example in `CONTRIBUTION_CHANNELS.md`. Great fix for a real observability gap.

Before merge:

## 🔴 Critical

1. **Rebase onto current main.** The diff shows 7 files changed, but only 3 are from commits on this branch (`6288da8`, `28f1ef2`). The other 4 (`behaviors/agents.yaml`, two files under `experiments/`, `tests/test_dicts.py`) are stale from before PRs #180 and #181 updated them on main. Under squash merge, the branch tip wins — which would silently revert the `amplifier-bundle-skills` → `amplifier-module-tool-skills` URL migration those PRs landed, with no CI signal. Rebase resolves this regardless of merge strategy.

2. **Drop commit `28f1ef2` (BUNDLE_GUIDE.md `did_mount()` section) and land it in a follow-up PR after amplifier-core #63 merges.** Two concrete reasons:

   a. **Cross-reference is broken.** Line 766 points to `core:docs/contracts/CONTRACTS.md § did_mount` — that path doesn't exist. The actual file is `core:CONTRACTS.md` (repo root); `docs/contracts/` contains per-protocol files, no `CONTRACTS.md`.

   b. **Content will be stale.** amplifier-core #63 is under REQUEST CHANGES with items that directly affect this section — possible rename (`did_mount` → `post_compose` or `on_session_ready`), clarification that it fires per-session (not process-lifetime), dispatch ordering, `module:did_mount_failed` failure event. Worse: if this PR lands before core #63, the docs instruct authors to implement a method that silently no-ops in the released core — no error, no log, just a method that never fires. That's a support-load bomb.

   The tool-delegate migration (`6288da8`) doesn't depend on `did_mount` and can land independently.

## 🟡 Important

3. **Disclose the `_spawn_new_session` extraction and `isinstance` guard in the PR body / commit message.** The 114-line `__init__.py` diff is larger than "migrate one pattern" suggests because it bundles three legitimate changes: (a) the stated migration, (b) extraction of `_spawn_new_session()` for testability, (c) `isinstance(orch_section, dict)` guard — a real bug fix when orchestrator config is a bare string like `"loop-basic"` causing `AttributeError` on `.get("config")`. All look correct; they just aren't mentioned. Worth a sentence for future `git blame`, and the `isinstance` fix deserves a regression test.

4. **Log the swallowed exception** at `__init__.py:914-919`:
   ```python
   except Exception as e:
       logger.warning("Failed to resolve agents config, defaulting to empty: %s", e)
       agents = {}
   ```
   Silent swallows produce misleading downstream "Agent not found" errors when the real failure was upstream.

5. **Pin `amplifier-core` minimum version** in `pyproject.toml`. Current constraint `amplifier-core>=0.1.0` allows installation with a core that predates `register_contributor`, giving users `AttributeError` at mount time. Bump to the release where the method was introduced.

## 🟢 Nit

6. `scripts/lint_observability.py:157` — the docstring cites tool-delegate as an example of the old pattern. After this PR, tool-delegate is the counter-example. Remove "as used in tool-delegate" from the comment.

---

Happy to pair on any of these. Once rebased and the docs commit is pulled, the migration itself is in great shape.

colombod:
All items addressed:

**Critical 1 — Rebase:** Rebased onto `origin/main`. The 4 stale files (`behaviors/agents.yaml`, `experiments/`, `tests/test_dicts.py`) are resolved — no revert risk.

**Critical 2 — Drop BUNDLE_GUIDE.md:** Dropped. The `on_session_ready()` docs will land as a separate PR after amplifier-core #63 merges. PR now contains only `modules/tool-delegate/` changes.

**Important 3 — Disclose bundled changes:** Commit message updated to explicitly name the `_spawn_new_session()` extraction and the `isinstance(orch_section, dict)` guard. Added `tests/test_delegate_spawn_new_session.py` with 3 regression tests covering: bare-string orchestrator config (no crash), `None` config forwarding, and the normal dict path.

**Important 4 — Log swallowed exception:** The handler at line 920 now logs `exc_info=True` so the root cause surfaces instead of just the downstream "Agent not found" error.

**Important 5 — Pin amplifier-core version:** `amplifier-core>=0.1.0` → `>=1.0.0` in `pyproject.toml`.

**Nitpick 6 — lint_observability.py docstring:** Line 157 updated — tool-delegate is now cited as the counter-example of the old pattern, not an example of it.

Tests: 37/37 passing (includes 3 new spawn regression tests).

bkrabach (approved):
All 6 asks from the previous round are resolved. Clean commit, clean rebase, full regression suite passes (1248 tests).

One small tightening before merge:

**Version pin is still too loose.** `amplifier-core>=1.0.0` in `pyproject.toml` allows installation against v1.0.x / v1.1.x, which predate `register_contributor`. I verified the method was added in the v1.2.x line (present in v1.2.1 through v1.3.3). Please bump the pin to `>=1.2.1` (or whichever earlier v1.2.x introduced the method — I didn't bisect).

Otherwise: approved.

colombod:
Version pin bumped to `amplifier-core>=1.2.1` (commit amended, force-pushed). Verified: `register_contributor` is absent in v1.1.1 and present in v1.3.3; your v1.2.x finding is consistent with the evidence.

bkrabach (approved):
Verified `amplifier-core>=1.2.1` pin in `modules/tool-delegate/pyproject.toml`. Last concern resolved — ready to merge.

PR #183, opened 2026-04-23
docs(bundle-guide): document on_session_ready() lifecycle method

colombod:
## Summary

Adds `on_session_ready()` documentation to `BUNDLE_GUIDE.md` as an available module lifecycle method.

> **Depends on:** microsoft/amplifier-core#63 — do not merge until that PR lands and a new amplifier-core release is published to PyPI.

## What this adds

A new subsection in `BUNDLE_GUIDE.md` immediately after the `mount()` documentation covering:

- **Contract** — async-only, coordinator-only signature, non-fatal exceptions, return value ignored
- **When to use** — cross-module event subscription, capability wiring after full composition, defensive dual-path elimination
- **Before/after example** — dual-path mount() pattern vs clean single-path on_session_ready()
- **Session-scoped pattern** — using `coordinator.register_capability` / `get_capability` instead of process-scoped globals

## Why gated on amplifier-core#63

The documented method silently no-ops on any installed amplifier-core that predates the release. Landing docs before the code ships creates a support-load trap. This PR should be merged in the same release cycle as the core change.

### DIFF
--- a/docs/BUNDLE_GUIDE.md
+++ b/docs/BUNDLE_GUIDE.md
@@ -734,6 +734,38 @@ async def mount(coordinator: Any, config: dict[str, Any] | None = None) -> dict[
     return {"name": "tool-my-tool", "version": "0.1.0", "provides": ["my_tool"]}
 ```
 
+### `async def on_session_ready(coordinator) → None` (optional)
+
+Called once, after every module across all phases has completed `mount()`. The coordinator is fully composed at this point: all providers, tools, hooks, capabilities, and contributions are registered and visible. Return value is ignored. Exceptions are non-fatal — logged as warnings, remaining `on_session_ready()` calls continue.
+
+**When to use it:**
+- You need to discover what other modules registered (events, capabilities, contributions) and can't know that at `mount()` time
+- You need to wire your module to another module's capability and want a guarantee it's present
+- You're replacing a defensive "check if capability exists yet" pattern with a clean single path
+
+**Example — eliminating a defensive dual-path:**
+
+```python
+# Before — defensive dual-path in mount() (register handler AND check immediately,
+# because mount order determines whether the capability exists yet)
+async def mount(coordinator, config):
+    coordinator.hooks.register("skills:discovered", on_skills_discovered)
+    # Also check immediately in case tool-skills already ran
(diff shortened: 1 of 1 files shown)


bkrabach (changes requested):
Content is clean — rename is consistent, cross-reference path correct, contract summary matches CONTRACTS.md, examples use real coordinator API, and the PR body correctly declares both the core #63 dependency and the "don't merge until a new PyPI release" gate.

One blocker:

🔴 **Rebase needed — same branch-staleness issue as #182 round 1.** `git diff main pr-183 --stat` shows 5 files changed, but the commit only touches `BUNDLE_GUIDE.md`. The other 4 files (`behaviors/agents.yaml`, two under `experiments/`, `tests/test_dicts.py`) appear because the branch was cut before PRs #180/#181 landed on main. Under squash merge, the branch tip wins — silently reverting the `amplifier-bundle-skills` → `amplifier-module-tool-skills` URL migration. Rebase resolves it.

🟢 Polish (non-blocking):
The Before/After labels over-claim vs what the examples show. "Before" is labeled "process-scoped global (anti-pattern)" but doesn't include a `global` declaration; "After" is labeled "session-scoped" but doesn't use `register_capability`. Either mirror CONTRACTS.md § on_session_ready (which has both) or soften the labels to describe the dual-path → single-path simplification the current snippet actually demonstrates.

colombod:
Rebased onto origin/main (stale files resolved — branch now touches only BUNDLE_GUIDE.md).

Polish: Before/After labels corrected. The comment lines now read:
- Before: 'defensive dual-path in mount() (register handler AND check immediately, because mount order determines whether the capability exists yet)'
- After: 'single path in on_session_ready() (all modules mounted, no ordering assumption)'

Labels now describe what the snippet actually demonstrates rather than over-claiming 'process-scoped global' or 'session-scoped'.

bkrabach (approved):
Rebased cleanly (branch now touches only `BUNDLE_GUIDE.md`) and labels are polished — "defensive dual-path in mount()" vs "single path in on_session_ready()" accurately describe what the example demonstrates. Cross-ref path is correct.

Approving. Per PR body's own gate, this lands after core #63 merges and a new amplifier-core release publishes to PyPI.

PR #196, opened 2026-05-09
fix(build-up): disable self-delegation at parent level + close three model-facing leaks

bkrabach:
## Motivation

Forensic analysis of an eval-matrix outlier (Qwen3.6-35B-A3B on build-up × fix-failing-test, mean 215 LLM requests vs 29-34 for cloud models) traced a **30-minute timeout to runaway `delegate(agent='self')` loops**. 

While the tool-delegate module has runtime enforcement to reject self-delegation when disabled, three separate leaks still taught the model that `'self'` was a valid target:

1. **delegation-mechanics.md** — bundle context explicitly taught `agent='self'` with four worked examples
2. **input_schema parameter description** — static, unconditionally listed `'self'` as an example
3. **'No agents registered' fallback message** — also mentioned `agent='self'`

## Changes

| File | Change | Rationale |
|---|---|---|
| `experiments/build-up/behaviors/build-up-foundation.yaml` | Set `features.self_delegation.enabled: false` for parent | Parent has no real tools; self-spawn is useless. Sub-agents keep enabled via independent tool-delegate declarations. |
| `experiments/build-up/context/delegation-mechanics.md` | Removed all four `agent='self'` references; added upfront note explaining disabled state | No longer teaches pattern that runtime will reject. Updated code examples to use specialist names. |
| `modules/tool-delegate/amplifier_module_tool_delegate/__init__.py` | Made input_schema agent description dynamic; gated fallback message on flag | Appends enabled/disabled status note to schema. Fallback no longer mentions `'self'` when disabled. |
| `modules/tool-delegate/tests/test_delegate_self_delegation_disabled.py` | **New file: 8 tests** covering disabled path | Previously had zero test coverage for this code path. |

## Test Plan

✅ All 8 new tests pass  
✅ Full delegate test suite (45 tests total) passes — no regressions  
✅ End-to-end fix verified: `delegate(agent='self')` now returns sub-second ToolResult error instead of 30-minute timeout loop

## Architecture Note

Per-sub-agent override is already handled correctly: each sub-agent (build-up:explorer/planner/coder/tester) declares its **own independent** `tool-delegate` instance in its .md frontmatter. The parent's `settings.exclude_tools: [tool-delegate]` strips the parent's tool from inheritance; each sub-agent's declaration is honored with its own config. Sub-agents keep self-delegation enabled by default (part of their explorer scale-out pattern). No new config schema introduced.

Generated with Amplifier

### DIFF
--- a/experiments/build-up/behaviors/build-up-foundation.yaml
+++ b/experiments/build-up/behaviors/build-up-foundation.yaml
@@ -83,7 +83,19 @@ tools:
     config:
       features:
         self_delegation:
-          enabled: true
+          # Disabled at the parent level so the orchestrator cannot delegate
+          # to itself. The build-up parent has only `todo` + `delegate` —
+          # spawning another instance of the parent bundle (which is what
+          # `agent="self"` does) produces a sub-instance with no real tools
+          # (no bash/filesystem/search/edit_file) that can only thrash mode/
+          # todo until context overflows.
+          #
+          # Sub-agents (build-up:explorer, :planner, :coder, :tester) declare
+          # their own `tool-delegate` instances independently in their .md
+          # frontmatter. They keep self-delegation enabled (the default) for
+          # legitimate parallel-dispatch patterns at the specialist layer
+          # — see explorer.md's parallel-dispatch guidance.
+          enabled: false
         session_resume:
           enabled: true
         context_inheritance:
--- a/modules/tool-delegate/amplifier_module_tool_delegate/__init__.py
+++ b/modules/tool-delegate/amplifier_module_tool_delegate/__init__.py
@@ -259,13 +259,52 @@ Agent usage notes:
 
         Returns:
             JSON schema for the tool input with structured parameters
+
+        Implementation note:
+            The schema body is kept as a pure literal in ``_static_input_schema()``
+            so foundation's static token-cost estimator
+            (``amplifier_foundation.bundle_docs.tool_schema._extract_input_schema``)
+            can extract it via ``ast.literal_eval``. The dynamic note about
+            whether self-delegation is enabled in this session is appended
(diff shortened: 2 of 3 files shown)


bkrabach:
Course correction per user feedback: reverted the `description` property's "no agents registered" fallback gating (kept that string generic). The bundle-specific guidance about self-delegation being disabled lives in `delegation-mechanics.md` (change #2 in this PR). The dynamic `input_schema` agent-description note is unchanged — the user explicitly preferred that approach.

Test file refactored to assert specifically on the feature-listing line, not on any occurrence of `agent="self"` in the description. Tests still 8/8 + 45/45 passing.

Amended commit: `e6768d6`

bkrabach:
Per user feedback, fully reverted the changes to `experiments/build-up/context/delegation-mechanics.md` — keeping that file as it is on main. The PR now narrows to: (a) the parent's `self_delegation.enabled: false` flag flip, (b) the dynamic `input_schema` status note, and (c) the new tests. The runtime enforcement at `__init__.py:812-816` (which already exists on main) is what actually rejects `agent="self"` calls; this PR adds the operator-facing config flip and the model-facing schema signal that the call will be rejected.

Tests still 8/8 + 45/45 passing.

**New commit:** `37108e6`

bkrabach:
CI fix: foundation's static token-cost estimator (`tool_schema._extract_input_schema`) uses a regex+`ast.literal_eval` pipeline that cannot evaluate string concatenation in the dict literal returned by `input_schema`. Refactored to keep the static literal in a separate helper `_static_input_schema()` and apply the dynamic self-delegation status note as a post-construction mutation in the `input_schema` property. All 46 affected tests pass locally; behaviorally identical at runtime.

PR #199, opened 2026-05-11
fix(cost): replace per-child contributors with capability-attached aggregator

kenotron-ms:
## Summary

Fix multi-turn cost compounding in `bridge_child_cost`. Previously, each call registered a per-child contributor via `register_contributor("session.cost", f"delegate:{child_session_id}", ...)` — and since `register_contributor` always appends with no dedup by name, multi-turn resumes of the same child added a new contributor each turn. After N resumes: N entries each holding the cumulative cost → N× over-count.

## Fix

Replace per-child contributors with a single `"delegate-children"` contributor per parent coordinator, backed by a mutable dict attached to the coordinator via `register_capability`. First bridge call registers the contributor and attaches the dict; subsequent calls find the existing dict via `get_capability` and mutate it in place — replace-on-dup semantics across multi-turn resumes, per-parent isolation by construction.

State lives and dies with the coordinator — no module-level dict, no `id()` reuse hazard.

- No Rust changes required
- Replace-on-dup semantics for multi-turn resumes
- Multiple distinct children summed in one contributor
- Per-parent isolation: each coordinator has its own state, independent of memory address reuse

## Tests

- `test_bridge_child_cost_no_compounding_on_multi_turn_resume` — 5 resumes of the same child → 1 register call, contributor returns latest value
- `test_bridge_child_cost_multiple_children_sum_correctly` — 3 distinct children → 1 register call, contributor returns the sum
- `test_bridge_child_cost_isolates_state_per_parent` — two distinct parents bridging children with the same id → each parent sees only its own
- Updated `test_prepared_bundle_spawn_bridges_child_cost` for the new contributor name and capability-attached state

## Validation

- Foundation test suite: cost bridge path 6/6 pass. No regression in any test that wasn't already failing on `main` (macOS path-normalization failures, unrelated).
- End-to-end validated in an isolated container: real multi-turn delegate session with the same child session_id resumed across multiple turns produces a single `"delegate-children"` contributor whose value matches the per-message cost sum exactly — no compounding observed.

### DIFF
--- a/amplifier_foundation/bundle/_prepared.py
+++ b/amplifier_foundation/bundle/_prepared.py
@@ -26,6 +26,14 @@ from amplifier_foundation.bundle._dataclass import Bundle
 
 logger = logging.getLogger(__name__)
 
+# Coordinator capability key for bridged delegate costs.
+# Stored value: dict[child_session_id, Decimal]. Enables replace-on-dup semantics:
+# re-bridging the same child (multi-turn resume) updates its slot rather than
+# appending a new contributor each time. State lives on the coordinator itself
+# via register_capability, so it's bounded by coordinator lifetime — no module-
+# level state, no id() reuse hazard.
+_DELEGATE_COSTS_CAPABILITY = "_bridge_delegate_costs"
+
 
 # Collects cost_usd contributions from a session.cost channel and returns the
 # total as Decimal, or None when no cost data is present (e.g. self-hosted models).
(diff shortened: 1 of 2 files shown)


kenotron-ms:
Closing — the bug this PR targets does not manifest end-to-end on main.

The PR's premise was that `bridge_child_cost` over-counts when a child session is resumed multiple times: each resume calls `register_contributor("session.cost", f"delegate:{child_session_id}", ...)`, and since the kernel's `register_contributor` always appends with no dedup, N resumes produce N entries.

The append behavior is real. The over-count it would cause is not — pre-rework DTU measurement against a real multi-turn delegate flow confirms the predicted N× compounding does not occur end-to-end.

**Why the bug doesn't manifest:** child coordinators are FRESH on every spawn/resume call. `AmplifierSession.__init__` unconditionally creates a new `ModuleCoordinator` — there's no path that reuses the prior coordinator on resume. So the N contributors registered across N resumes hold N PER-TURN costs, not N copies of the cumulative. `sum([turn-1 cost, turn-2 cost, ...])` is the correct total by structural accident.

The rework is architecturally cleaner than per-call `register_contributor` and would prevent the bug if coordinator lifetime ever changed to persistent-across-resumes. Closing per "today's needs only" — re-derive from this PR's history if/when that future arrives.

PR #213, opened 2026-05-18
feat(mentions): emit mentions:resolved event after each resolution pass

colombod:
## Problem

The `@mention` resolver is completely silent. When `PreparedBundle` loads context files declared in the `context:` YAML section, nothing observable happens. The resolved content lands inside multi-KB XML blobs in `session:start.data.raw` that are:

- Redacted in ~65% of sessions by `hooks-redaction`
- Absent entirely for runtime (per-turn) @mentions
- Fragile to parse — the XML format is a display convention, not a schema contract

This creates three concrete gaps:

| Gap | Impact |
|-----|--------|
| **No resolution visibility** | Cannot determine which @mentions resolved vs. failed, or which files were loaded |
| **Silent failures** | A bundle declares a context file that no longer exists — nobody finds out |
| **No turn-level tracking** | User types `@AGENTS.md` mid-session — invisible to hooks and observers |

---

## Two-layer event model

This PR introduces `mentions:resolved` as a **foundation-layer event** — the per-mention parse fact. Bundle attribution and content-inclusion facts belong at a different layer:

| Event | Layer | Fires when |
|---|---|---|
| `context:include` | kernel constant (added `fd1feeb`, *"Context file inclusions (for debugging context size)"*) | A context file is added to the system prompt — any source |
| `mentions:resolved` | foundation/mentions | An `@`-syntax token is parsed and resolved |

These are distinct events at distinct layers. A user typing `@AGENTS.md` produces both: `mentions:resolved` (the parse fact) and `context:include` (the inclusion fact). Consumers needing the joined view assemble it from both. This PR covers `mentions:resolved` only.

---

## Solution

A single new event, `mentions:resolved`, emitted after each resolution pass inside the system-prompt factory. The event carries the resolution manifest — what was asked for, what resolved, what failed, what was deduplicated.

### Event schema

```python
{
  "source": "bundle_context",     # or "user_prompt" (app-layer, follow-up PR)

  "resolutions": [
    {
      "mention":       str | None,  # "@foundation:context/foo.md" or None for context: decls
      "resolved_path": str,         # absolute path on disk
      "source_type":   str,         # how the path was referenced:
                                    # "bundle_namespace" | "bundle_context_decl" |
                                    # "user_shortcut" | "project_shortcut" |
                                    # "home_shortcut" | "relative_path"
      "content_hash":  str,         # SHA-256 matching the deduplicator
      "is_new":        bool,        # True = first time this file appears in the session
    }
  ],

  "failed": [{"mention": str, "reason": str}],
  "deduplicated_count": int,
  "turn": None,                     # null for bundle_context; int for user_prompt
}
```

### Silent-fail policy

The `@AGENTS.md`-in-many-locations pattern is intentional — those non-resolutions are not failures and do not appear in `failed[]`. Only entries with an explicit `failure_reason` (`"not_found"`, `"permission_error"`) surface. Two specific cases are deferred to follow-up PRs:

- Bundle `context:` YAML declaration → file missing: validate at load time, surface via `/config show context`. Separate PR.
- User-typed `@mention` → zero resolutions (typo): one-line UX warning. Separate PR.

### Where it fires

```
session:start
  ├─ [turn 1]
  │     execution:start
  │     mentions:resolved  ← source=bundle_context (all context files)
  │     provider:request
  │
  ├─ [turn 2, same files]
  │     execution:start
  │     (no event — deduplication guard)
  │     provider:request
  │
  └─ session:end
```

---

### Hook subscription: two complementary mechanisms

| Mechanism | Path | Effect |
|---|---|---|
| `inject_additional_events(FOUNDATION_OBSERVABILITY_EVENTS)` | mount plan → hook `additional_events` config | hooks-logging and hook-context-intelligence subscribe at session init |
| `coordinator.register_contributor("observability.events", ...)` | coordinator contribution channel | event discoverable via `collect_contributions("observability.events")` at runtime |

Both are registered in `create_session()` and `spawn()`, guarded by `if bundle.instruction or bundle.context`. Empty-bundle sessions do not register — they never emit the event.

---

### `amplifier_foundation/mentions/models.py`

Added `failure_reason: str | None = None` to `MentionResult`. Defaults to `None` — all existing call sites are unaffected.

### `amplifier_foundation/mentions/loader.py`

`_resolve_mention()` now sets `failure_reason` at each failure branch (`"not_found"` | `"permission_error"`).

### `amplifier_foundation/bundle/_observability.py`

Added `"mentions:resolved"` to `FOUNDATION_OBSERVABILITY_EVENTS`.

### `amplifier_foundation/bundle/_prepared.py`

**Added:**
- `_determine_source_type(mention) -> str` — classifies the syntax form of a mention string

**Updated:**
- `_build_resolutions` — emits `{mention, resolved_path, source_type, content_hash, is_new}` per entry
- `_create_system_prompt_factory` — emits event; `failed[]` filter uses `mr.failure_reason is not None` so PermissionError cases surface
- `create_session()` — registers on `observability.events` inside the bundle-content guard
- `spawn()` — same registration, same guard; `spawn()` bypasses `create_session()` so both sites are required

---

## Tests — 41 passing, 0 warnings

`tests/test_mentions_resolved_event.py`

| Group | Count | What it proves |
|---|---|---|
| `TestMentionResultFailureReason` | 5 | New field, backward-compatible |
| `TestLoaderFailureReason` | 4 | `failure_reason` set at every failure site |
| `TestDetermineSourceType` | 11 | Syntax-label classifier, all mention forms |
| `TestBuildResolutions` | 4 | Payload shape, `is_new`, `resolved_path` absolute |
| `TestFactoryEmission` | 6 | Turn 1 emits; turn 2 silent; new file re-emits; PermissionError in `failed[]` |
| `TestReviewerFindings` | 3 | `resolved_path` absolute, `spawn()` registers, PermissionError in `failed[]` |
| `TestObservabilityRegistration` | 5 | Both `create_session()` and `spawn()` register; empty bundle does not |

---

## What is NOT in this PR

- **`user_prompt` source (Path B)** — `amplifier_app_cli/main.py:_process_runtime_mentions()`. Schema and registration established here; app-layer emission is a follow-up PR
- **`context:include` emission** — emitting the kernel-layer inclusion fact is a separate PR; the constant and its semantics are now documented
- **Bundle `context:` YAML missing file** — surface via `/config show context` with a `[missing]` flag; separate PR
- **User-typed `@mention` typo warning** — one-line UX surface; separate PR
- **`ContextFile.origins` field** — tagging attribution on the dataclass itself; separate PR

### DIFF
--- a/amplifier_foundation/bundle/_observability.py
+++ b/amplifier_foundation/bundle/_observability.py
@@ -26,6 +26,7 @@ _DEFAULT_SUBSCRIBER_MODULES: frozenset[str] = frozenset(
 # Rust kernel's ALL_EVENTS list.
 FOUNDATION_OBSERVABILITY_EVENTS: tuple[str, ...] = (
     "session:config",  # emitted by _session_exec.emit_raw_field_if_configured (PR #79)
+    "mentions:resolved",  # emitted by PreparedBundle system-prompt factory (PR #213)
 )
 
 
--- a/amplifier_foundation/bundle/_prepared.py
+++ b/amplifier_foundation/bundle/_prepared.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 
 import asyncio
 import logging
-import re
 from dataclasses import dataclass
 from dataclasses import field
 from decimal import Decimal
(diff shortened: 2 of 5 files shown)


bkrabach (commented):
Acknowledging both commits — the original `21b543c` and the `Bundle.origins` refactor in `ee6ad95`. The second commit is the right direction; replacing the cache-path regex with the existing provenance machinery is architecturally aligned with `/config show` and inherits the via-behavior chain for free.

Five things to land before this is mergeable. Two design calls + three mechanical. The design calls come from our call recording — making them explicit here so future readers don't need to know what was said.

### Design call 1: Two-layer event model

`context:include` (kernel) and `mention:resolved` (foundation) are distinct events at distinct layers. The kernel-level constant is intentional — added in `amplifier-core@fd1feeb` (2025-10-13) for *"Context file inclusions (for debugging context size)"* — and is the right home for the file-inclusion fact:

| Event | Layer | Fires when | Bundle attribution |
|---|---|---|---|
| `context:include` | kernel constant, emitted from foundation | A context file is added to the system prompt (any source: bundle YAML, mention resolution, app-cli MentionLoader) | Carries `bundle` and/or `origins` when known |
| `mention:resolved` | foundation/mentions | An `@`-syntax token is parsed and resolved | No bundle awareness — the mention string IS the convention; consumers join elsewhere |

A bundle's `context:` YAML declaration → `context:include` only. A user typing `@AGENTS.md` mid-session → `mention:resolved` (the parse fact) and `context:include` (the inclusion fact). Consumers needing the joined view assemble it from the two events plus `/config show` data.

Concretely for this PR: **drop `origins` from the `mention:resolved` payload** per what we agreed in the call ("the bundle part in the payload of the event is not really needed because the mention contains... `app foundation column context`"). The `Bundle.origins` work was the right source for attribution — but attribution belongs on the `context:include` side, not the per-mention side. The per-mention event stays small and at-mention-specific.

### Design call 2: Silent-fail policy

The `@AGENTS.md`-in-many-locations pattern is intentional. Failures there are non-events. Two specific cases deserve different treatment, neither of which `mentions:resolved` should try to absorb on its own:

- **Bundle `context:` YAML declaration → file doesn't exist.** Validate at load time. Hard error, or `[missing]` flag visible in `/config show context`. This is a bundle-author bug; surfacing at runtime instead of build-time is the actual concern. Not a `mentions:resolved` failure path.
- **Runtime `@mention` typed by user → zero resolutions.** Keep silent for `if-exists` cases (e.g., a bundle including `@AGENTS.md` across five potential paths). Surface a one-line user-visible warning when a *user-typed* mention has zero resolutions (typo case). Preserves the AGENTS.md pattern; adds UX for user mistakes.

Both of these land as separate follow-up PRs after this one. Flagging here so the design is documented and scope on #213 stays tight.

### Mechanical fixes still pending

1. **`spawn()` is missing `register_contributor`.** The PR description lists this as one of three post-review corrections, but `_prepared.py:694-900` (the `spawn()` body) shows no such call — verified by grep. Add the `register_contributor("observability.events", "foundation:mention-resolver", lambda: [_MENTIONS_RESOLVED_EVENT])` call after `child_session.initialize()`, mirroring the `create_session()` site. Add a test exercising `spawn()` (not `create_session()`) that fails without it — none of the three current `TestObservabilityRegistration` tests cover this path.

2. **`failure_reason` filter is dead at 4 of 5 sites.** The filter `if not mr.found and mr.resolved_path is None` at `_prepared.py:476` excludes every case where the path was successfully resolved but the read failed (PermissionError on file, OSError on file, etc.). The four `failure_reason` values set in `loader.py:157-183` for those cases never reach the `failed[]` payload. Either change the filter to `if not mr.found and mr.failure_reason is not None`, or drop `failure_reason` from the four unreachable sites since it's currently dead code.

3. **`register_contributor` is unconditional in `create_session()`.** Called regardless of whether the bundle has instruction/context. Sessions with empty bundles advertise an event that never fires. Move the call inside the same `if self.bundle.instruction or self.bundle.context or self.bundle._pending_context:` guard that gates the factory.

### Out of scope for this PR (separate work)

- **No-bundle `MentionLoader` observability** (`app-cli/lib/mention_loading/loader.py`). Sessions using mentions without bundles currently have zero observability. With the two-layer model above, that path emits `context:include` when it adds messages. Separate PR, in app-cli.
- **`/config show context` existence status.** Render `[missing]` for context files whose `path.exists()` is False. Needs an `exists: bool` field on `ItemRecord` and a renderer tweak. Separate PR, in foundation.
- **PR description correction.** The "three post-review corrections" claim is stale: test count is 28, not 42; `TestReviewerFindings` class doesn't exist; the spawn() fix is absent. Update the description so future readers aren't misled.

colombod:
All five items addressed in `801f6c1`. Notes on each:

---

**Design call 1 — two-layer model**

Accepted. `_to_origins_key`, `_lookup_origins`, `bundle_origins` parameter, and the `origins` field are all gone from `mentions:resolved`. The schema is now:

```python
{
  "mention":       str | None,
  "resolved_path": str,         # absolute
  "source_type":   str,         # "bundle_namespace" | "bundle_context_decl" | ...
  "content_hash":  str,
  "is_new":        bool,
}
```

Bundle attribution will belong on `context:include`. `_determine_source_type` stays — it classifies the syntax form of the mention string, which is still useful data at the `mentions:resolved` layer (tells you how the path was expressed, not who introduced it).

---

**Design call 2 — silent-fail policy**

Noted and documented in the PR description. The two specific follow-up cases (bundle `context:` YAML file missing surfaced via `/config show`, user-typed `@mention` typo surfaced as a one-line UX warning) are out of scope for this PR and noted as separate follow-up work.

---

**Mechanical fix 1 — `failure_reason` filter**

Changed from `mr.resolved_path is None` to `getattr(mr, "failure_reason", None) is not None`. Now only entries with an explicitly set `failure_reason` surface in `failed[]`. The `test_permission_error_on_file_appears_in_failed_list` test in `TestReviewerFindings` exercises this path end-to-end.

---

**Mechanical fix 2 — `spawn()` missing `register_contributor`**

Added after `await child_session.initialize()`, guarded by `if effective_bundle.instruction or effective_bundle.context:` mirroring the `create_session()` guard. New test `test_spawn_registers_when_child_has_content` in `TestObservabilityRegistration` covers this path specifically (uses a child bundle with an instruction so the guard fires).

---

**Mechanical fix 3 — unconditional `register_contributor`**

Moved inside the `if self.bundle.instruction or self.bundle.context or self.bundle._pending_context:` guard. New test `test_empty_bundle_does_not_register` asserts the channel is not registered for an empty bundle.

---

On `OQ-2 / context:include` semantics — thank you for the authoritative reference (`fd1feeb`, *"Context file inclusions (for debugging context size)"*). Updating the PR description to close that open question with the definitive answer: distinct events at distinct layers, both intentional.

bkrabach (approved):
All five items resolved in `801f6c1`. Approving.

- **Two-layer event model:** `origins` dropped from `mentions:resolved` payload; `_to_origins_key`, `_lookup_origins`, `bundle_origins` param all gone. Schema is now `{mention, resolved_path, source_type, content_hash, is_new}` — small, at-mention-specific. `_determine_source_type` correctly retained as syntax classification.
- **`spawn()` `register_contributor`:** `_prepared.py:170`, after `child_session.initialize()`, guarded by the same `effective_bundle.instruction or effective_bundle.context` check that gates the factory.
- **`create_session()` registration guard:** Moved inside the `if self.bundle.instruction or self.bundle.context or self.bundle._pending_context:` block at L122-124.
- **`failure_reason` filter:** Now `if not mr.found and getattr(mr, "failure_reason", None) is not None`. The defensive `getattr` for forward-compat with older installed stubs is a nice touch.
- **PR description:** Updated with the two-layer event model + silent-fail policy sections. `TestReviewerFindings` class now actually populated (was stale in earlier passes).

Follow-ups for `/config show` existence flag, no-bundle `MentionLoader` observability, and the actual `context:include` foundation-side wire-up tracked separately. None of those block this merge.

PR #268, opened 2026-07-06
fix(modules): run uv installs async to stop freezing the event loop

singh2:
## Summary

- **Durable root-cause fix** for a class of "synchronous subprocess freezes the resolver event loop" bugs. `ModuleActivator._install_dependencies` ran `subprocess.run(["uv","pip","install",...])` **synchronously on the asyncio event-loop thread, unbounded**, at both install sites (pyproject and requirements).
- A synchronous subprocess blocks the *entire* loop, so any slow install froze **all** coroutines on it — **including liveness heartbeats**. In the Amplifier Resolve understudy resolver this caused workers to be reaped mid-bring-up (**~25% of runs**) during heavy bundle/module installs. `asyncio.gather` in `activate_all` cannot parallelize a synchronous subprocess: the first coroutine to hit it stalls everything behind it.

## Root cause

Both install call sites used `subprocess.run(..., check=True, capture_output=True, text=True)` with **no timeout**. Running on the loop thread, a single wedged `uv` install:
1. blocks every other coroutine on the loop (heartbeats included), and
2. defeats `asyncio.gather` — installs that *look* concurrent are serialized behind the first blocking call, with no deadline to break a hang.

## The fix

**1. Non-blocking installs with a wall-clock bound.** Both `subprocess.run` sites are replaced by an async `_run_install` helper built on `asyncio.create_subprocess_exec` + `asyncio.wait_for(communicate(), timeout=DEFAULT_INSTALL_TIMEOUT=300s)`. On timeout the child is **killed and reaped** (no zombie) and a typed `ModuleInstallTimeout` is raised — fail loud, never hang. A timeout can now actually fire, because the loop is no longer frozen inside a synchronous `subprocess.run()`.

The prior error surface is **preserved**:
- non-zero exit → `subprocess.CalledProcessError` carrying decoded stdout/stderr (as `capture_output=True, text=True` gave), and
- missing `uv` binary → `FileNotFoundError`.

**2. Serialize installs with a per-event-loop lock (and why it avoids a *new* race).** Making installs non-blocking means `activate_all()`'s `asyncio.gather()` can, for the first time, run multiple editable `uv pip install` invocations **truly concurrently into the same interpreter environment**. Concurrent editable installs race on shared site-packages metadata (`.pth`, `RECORD`, `dist-info`). To close that newly-opened race, the install + state-mark critical section is serialized with an `asyncio.Lock` that is:
- created **lazily inside the running loop** (never binds a loop at import time), and
- held in a `WeakKeyDictionary` **keyed by the running loop** — so each child-session loop foundation spawns gets its own lock, avoiding "bound to a different event loop" errors, and entries disappear when a loop is GC'd.

## Test plan

- [x] New `tests/test_activator_install.py` (7 tests): success path, non-zero-exit error surface, timeout kill+reap, and a **loop-not-blocked regression guard** proving a concurrent task makes progress while a slow install runs.
  ```
  $ uv run pytest tests/test_activator_install.py -q
  7 passed in 0.42s
  ```
- [x] Full suite: **301 passed, 9 failed**.
  ```
  $ uv run pytest -q
  9 failed, 301 passed in 0.77s
  ```
  The 9 failures are **pre-existing and unrelated** to this change: macOS `/var` → `/private/var` symlink assertions in `test_paths.py`, `test_registry.py`, and `test_sources.py` (none touch the activator). They reproduce identically on the clean base.

## Scope / rollout note

This is the **durable root-cause fix** for the "sync subprocess freezes the resolver loop" bug class. It is defense-in-depth alongside two other mitigations already in place for the Amplifier Resolve go-live:
- a **stack-level pre-bake** that removes the install trigger at bring-up time (removes the *trigger*), and
- the **backend liveness watchdog** as the safety net.

This PR removes the *mechanism* so the failure mode cannot recur through this code path.

Generated with [Amplifier](https://github.com/microsoft/amplifier)

### DIFF
--- a/amplifier_foundation/modules/activator.py
+++ b/amplifier_foundation/modules/activator.py
@@ -15,12 +15,104 @@ import logging
 import subprocess
 import sys
 from pathlib import Path
+from weakref import WeakKeyDictionary
 
 from amplifier_foundation.modules.install_state import InstallStateManager
 from amplifier_foundation.paths.resolution import get_amplifier_home
 from amplifier_foundation.sources.resolver import SimpleSourceResolver
 
 logger = logging.getLogger(__name__)
+
+# Wall-clock bound (seconds) for a single ``uv`` install invocation. This is only
+# meaningful because installs are now run via a non-blocking async subprocess:
+# a timeout can never fire while the event loop is frozen inside a synchronous
+# subprocess.run(). Generous enough for a from-source editable build; can be
+# lifted into config later if this module grows a config surface.
+DEFAULT_INSTALL_TIMEOUT = 300
+
+
+class ModuleInstallTimeout(RuntimeError):
+    """Raised when a module install exceeds DEFAULT_INSTALL_TIMEOUT seconds.
+
+    Fails loud instead of hanging: the previous synchronous ``subprocess.run()``
+    had no timeout, so a wedged ``uv`` install would freeze the entire event loop
+    (and any liveness heartbeat) until the worker was reaped.
+    """
+
+
+# Per-event-loop install lock registry.
+#
+# Making installs non-blocking means activate_all()'s ``asyncio.gather()`` can,
+# for the first time, run multiple editable ``uv pip install`` invocations
+# *truly* concurrently into the SAME interpreter environment. Concurrent editable
+# installs race on site-packages metadata (.pth files, RECORD, dist-info), so we
+# serialize the install critical section with an ``asyncio.Lock``.
+#
+# The lock is created lazily *inside the running loop* and keyed by that loop, so
+# it never binds to a loop at import time and never triggers cross-loop
+# "bound to a different event loop" errors when foundation spawns child sessions
+# that each drive their own event loop. Entries are weakly held: when a loop is
+# garbage-collected its lock disappears automatically.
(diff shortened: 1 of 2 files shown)


singh2:
Closing this PR **without merging** — parking the work, not discarding it.

**This is a valid, durable fix.** It correctly addresses a real defect: `ModuleActivator` runs a synchronous `uv pip install` that blocks the event loop until the subprocess returns, freezing all concurrent async work in the process. The change here is the right shape — run the install as an async subprocess, wrap it in a timeout so a hung install can't wedge the loop indefinitely, and guard it with a per-event-loop lock to serialize concurrent activations safely. Nothing about the diagnosis or the implementation is wrong.

**Why it's being parked now:** we currently cannot land changes into `amplifier-foundation`, and this fix is **not required for the Amplifier Resolve go-live**. A stack-level pre-bake mitigation in `amplifier-resolve` already removes the trigger for the event-loop-freeze this addresses (modules are activated ahead of time, so the blocking install doesn't occur on the hot path), and the backend liveness watchdog remains as the safety net. So the go-live gate is met without merging foundation changes.

**Branch is preserved** (`fix/activator-async-install`) so the commit is fully recoverable. A foundation maintainer is welcome to reopen or adopt this — the fix stands on its own merits and should land whenever foundation changes can be accepted again.

PR #302, opened 2026-08-12
fix(spawn): memoize delegate list_models() with single-flight TTL cache

### LINKED ISSUE #369: [amplifier-foundation] Per-spawn delegate model-pattern resolution re-queries provider list_models() with no caching — parallel spawns storm GET /v1/models
## Summary

Every child/delegate session spawned with `provider_preferences` performs a **fresh, un-cached `provider.list_models()` network call**. Parallel spawns therefore fan out into N simultaneous `GET /v1/models` requests against the provider API. With recipes that spawn delegates in parallel, this produces a request storm proportional to spawn parallelism, adding latency to every spawn and (behind a shared egress proxy) measurable connection-table pressure.

## Call chain (all on current `main`)

1. `amplifier_foundation/bundle/_prepared.py` — `spawn()` applies provider preferences before creating the child session:

   `child_mount_plan = await apply_provider_preferences_with_resolution(child_mount_plan, provider_preferences, ...)` (~line 780)

2. `amplifier_foundation/spawn_utils.py` — `apply_provider_preferences_with_resolution` → `resolve_model_pattern`, which, for any glob-pattern model hint, queries the provider live (~line 208):

   ```python
   if provider and hasattr(provider, "list_models"):
       models = await provider.list_models()
   ```

3. `amplifier-module-provider-anthropic` — `list_models()` issues `await self.client.models.list()` (`__init__.py` ~line 935) on **every invocation**: no memoization, no TTL, no cross-call reuse.

No caching exists at any of the three layers (foundation resolution, provider instance, coordinator). A burst of K parallel delegate spawns with a pattern hint = K concurrent identical `GET /v1/models` calls.

## Observed impact

In a multi-agent workflow that spawns delegates in parallel waves (~80 wide at peak):

- ~80 simultaneous `GET /v1/models` per wave — the list is identical for all of them.
- ~16,600 such requests observed in a single day of repeated multi-agent runs.
- Each spawn pays the full network round trip before the child session is created, so spawn latency scales with upstream API latency.
- Behind a shared egress/credential proxy, the storm kept hundreds of concurrent connections warm and contributed to exhausting the proxy's connection table.

(Impact figures are measurements from our environment; the per-spawn call itself is directly verifiable from the code above.)

## Suggested fixes (any one materially helps; both together are best)

1. **Foundation-side memoization** (the layer that owns the storm): cache the resolved pattern result (or the fetched model list) on the coordinator, keyed by `(provider_name, pattern)` with a short TTL (e.g. 60s). Parallel spawns in one wave then collapse to a single upstream call.
2. **Provider-side TTL cache**: memoize `list_models()` on the provider instance for a short window. This fixes it for all callers but only within one provider instance's lifetime.

Option 1 also keeps the per-spawn latency benefit even when the provider instance is recreated each spawn.

Model lists change rarely (on the order of weeks), so even a multi-minute TTL is functionally exact while cutting request volume by orders of magnitude.

danshapiro:
## Summary

Fixes https://github.com/microsoft/amplifier/issues/369

Every child/delegate spawn with a glob-pattern model hint resolved the pattern by calling `provider.list_models()` live — and `resolve_model_pattern` runs once per spawn. Parallel delegate waves therefore fired one identical `GET /v1/models` per spawned session (observed: ~80 concurrent calls per wave, ~16k/day in a multi-agent setup), adding a full network round trip to every spawn and pressuring shared egress connection tables.

## Change

Memoize the fetched model list per provider instance in `spawn_utils`:

- **TTL** of 60s (`LIST_MODELS_CACHE_TTL_SECONDS`) — model catalogs change rarely, so a short window is functionally exact while collapsing a whole wave into a single upstream call.
- **Single-flight**: concurrent resolutions against the same provider coalesce behind one async lock — the first task fetches, the rest await its result. N parallel spawns = 1 upstream call.
- Cache keyed by provider instance via weak references, so entries die with the provider; lock is rotated if the provider is driven from a different event loop.
- **Failures are never cached** — errors propagate exactly as an uncached call would.
- Non-pattern hints still short-circuit before any query, unchanged.

## Tests

6 new regression tests in `tests/test_spawn_utils.py` (`TestListModelsSingleFlightMemoization`):

- 50 concurrent resolutions → exactly 1 upstream fetch
- sequential resolutions reuse the cache within the TTL
- refetch after TTL expiry
- exact (non-glob) model names never fetch
- independent provider instances cached independently
- a failing fetch is not cached; the next resolution retries

Full suite: 1533 passed (`uv run pytest tests/ -q`, excluding the gRPC integration file).

## Notes

- No behavior change to resolution results, ordering, or failure semantics — only fetch frequency and concurrency.
- Providers could separately add their own `list_models()` caching, but that cannot coalesce a parallel wave the way this shared-resolution point does.

### DIFF
--- a/amplifier_foundation/spawn_utils.py
+++ b/amplifier_foundation/spawn_utils.py
@@ -12,10 +12,10 @@ from __future__ import annotations
 import asyncio
 import fnmatch
 import logging
+import os
 import time
 import weakref
-from dataclasses import dataclass
-from dataclasses import field
+from dataclasses import dataclass, field
 from typing import Any
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


bkrabach (approved):
Due-diligence review + hardening: (1) only non-empty list_models results are cached, so soft-failure providers (ollama's [] on outage) no longer poison resolution for the TTL — repro-verified recovery on the call after the blip; (2) the in-flight fetch is a shared task delivered once to all concurrent waiters (success AND failure), preventing serial per-waiter retry campaigns during outages; (3) waiter timeout with uncached fallthrough so one hung fetch cannot stall a delegate wave; (4) TTL/wait knobs env-configurable. Single-flight collapse verified intact (80-wave -> 1 upstream fetch). Full suite 1563 passed. Known residual documented in-code: a soft-failure provider returning a plausible non-empty degraded list (chat-completions single-model fallback) is indistinguishable at this layer — needs a provider-side fix, tracked separately.

PR #318, opened 2026-08-27
fix(spawn): sub-agent provider override must outrank a priority-0 primary

bkrabach:
## The bug

`_apply_single_override()` in `amplifier_foundation/spawn_utils.py` promotes an override's target provider instance to `priority=0` in the **child** mount plan, but never demotes any other instance already sitting at (or below) that priority. Provider resolution picks the lowest-priority-number instance, ties broken by a **stable sort** on declaration order.

When a user's primary provider is declared **first** in `providers:` at `priority: 0` — the natural "make this my default model" config — every sub-agent provider preference (role-resolved routing-matrix prefs *and* agent-frontmatter `provider_preferences`) ties the primary at `priority=0`. The stable sort then hands the win to whichever instance was declared first: the primary, silently, no matter which instance the override selected.

## Live evidence

- A spawn recorded `provider_preferences=[{model: gpt-5.6-terra}]` in its `delegate:agent_spawned` event, yet the sub-session actually ran `gpt-5.6-sol` (the primary).
- In a 5-run live eval, **100% of sub-agent LLM calls ran the primary**, regardless of the role/agent's declared preference.
- Confirmed in source, in a live DTU canary, and in production captures. Completely silent — no error, no log, no signal that the override was dropped.

## The fix

`_apply_single_override()` (`amplifier_foundation/spawn_utils.py`) now, after promoting the override target to `priority=0`, walks the rest of **that same child mount plan's** provider list and demotes any other instance whose priority ties or beats the target's, to strictly below it (`target_priority + 1`). This guarantees the overridden instance always **strictly** wins its own child's provider resolution — it can no longer merely tie for it.

Scope/safety:
- Only touches the child mount plan being built for a given spawn. `apply_provider_preferences()` / `apply_provider_preferences_with_resolution()` are the only two callers of `_apply_single_override()`, and both operate exclusively on a freshly-cloned **child** mount plan (see `bundle/_prepared.py::PreparedBundle.spawn()`). Root/parent session provider resolution is untouched.
- No public data shape changes — the `providers[].config.priority` field already exists and is already mutated by this function; we've just made the invariant ("the override target strictly wins") actually hold.
- Mechanism is provider-agnostic — it demotes based on priority values only, regardless of module type, so it applies identically whether the tie is between two instances of the same provider or two different providers entirely.
- Emits a `logger.debug(...)` note whenever a tie-demotion actually fires, naming the demoted instance and the target it now strictly outranks, so this class of substitution failure is observable going forward instead of silent.

## Tests

Added `TestOverrideOutranksTiedPriorityZeroPrimary` to `tests/test_spawn_utils.py`, all of which **fail on unpatched main** and **pass with the fix**:

1. `test_override_second_instance_outranks_priority_zero_primary_same_type` — primary declared first at `priority=0`, a second instance of the *same* provider type at `priority=10`; override selects the second instance; asserts it strictly wins (and that the primary was demoted + a tie-break debug log fired).
2. `test_override_cross_provider_outranks_priority_zero_primary` — same shape but with two *different* provider types (openai primary, anthropic override target), proving the mechanism is provider-agnostic.
3. `test_override_selecting_primary_itself_no_demotion_needed` — override selects the primary itself; asserts no regression (secondary untouched, no spurious tie-break log).

Full suite: `uv run pytest tests/` → **1634 passed, 1 skipped** (no failures, no new lint/type issues introduced — verified pre/post-diff with `ruff`/`pyright`).

## Relationship to #302

Complements #302 (which memoized `list_models()` for glob-pattern resolution in this same file). No overlap in code paths touched — #302 lives in the model-list caching layer; this fix is entirely inside `_apply_single_override()`'s priority assignment.

Fixes: openai_improvement-ejq

### DIFF
--- a/amplifier_foundation/spawn_utils.py
+++ b/amplifier_foundation/spawn_utils.py
@@ -747,11 +747,16 @@ def _apply_single_override(
             are never overridden.
 
     Returns:
-        New mount plan with override applied.
+        New mount plan with override applied. The overridden (target)
+        provider is guaranteed to STRICTLY outrank every other provider in
+        this child mount plan's priority ordering (see the tie-break note
+        below) -- it is never merely tied with another instance.
     """
     # Clone mount plan and providers list
     new_plan = dict(mount_plan)
     new_providers = []
+
+    target_priority = 0
 
     for i, p in enumerate(providers):
         p_copy = dict(p)
@@ -764,7 +769,7 @@ def _apply_single_override(
                     if key not in PROTECTED_CONFIG_KEYS:
                         p_copy["config"][key] = value
             # Then enforce invariants — these always win
-            p_copy["config"]["priority"] = 0
+            p_copy["config"]["priority"] = target_priority
             p_copy["config"]["default_model"] = model
             logger.info(
                 "Provider preference applied: %s (priority=0, model=%s)",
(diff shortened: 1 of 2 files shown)


bkrabach:
**Maintainer admin-merge basis (documented per team policy):**

- **Author:** this PR was authored by our own automated account. GitHub structurally forbids self-approval, so no self-review can satisfy branch protection here — an external human reviewer is not in the loop for this change.
- **Test evidence:** demotes tied instances in the CHILD mount plan only (does not touch the primary session's mount plan). 3 new tests added — 2 fail-before/pass-after regression tests reproducing the silent all-sol substitution bug, 1 no-regression guard. Full repo suite: 1634 passed. Bug was proven in production eval captures (sub-agent provider override silently losing to a tied priority-0 primary).
- **CI status at merge time:** all 6 matrix jobs (ubuntu/windows × Python 3.11/3.12/3.13) + license/cla check green. See status checks on this PR.
- **Action:** merging via maintainer admin override (`gh pr merge --admin`) at the explicit direction of the repository owner, following the same sanctioned pattern previously used for gemini#39 (self-authored, self-approval-impossible, explicit maintainer direction).

Squash-merging now.

PR #319, opened 2026-08-27
fix(delegate): carry agent identity through resume completions + emit agent_resumed

bkrabach:
## What

Fixes delegate resume telemetry (work item `openai_improvement-9n6`).

When a delegation is **resumed** via `session_id` (the delegate tool's
resume path in `_resume_existing_session`), the parent stream got
`delegate:agent_completed` with no usable agent identity, and
`delegate:agent_resumed` was emitted without an `agent` field either.

Measured across 5 production eval runs:
- Fan-out counted from `agent_spawned` undercounted 18% (32 spawns vs
  39 delegate calls — the missing 7 were resumes).
- Completions on the resumed leg couldn't be attributed to an agent.
- Downstream context-intelligence graph ingestion mislabeled healthy
  resumed sub-sessions (see investigation notes below).

## Root cause

`DelegateTool._resume_existing_session()` emitted
`delegate:agent_resumed` and `delegate:agent_completed` with no
`"agent"` key at all. An agent name *was* derived from the session_id
(naive `split("_")[-1]`), but only **after both events had already
been emitted** — too late to matter, and never plumbed into either
payload.

## Fix

`modules/tool-delegate/amplifier_module_tool_delegate/__init__.py`:

- `DelegateTool.__init__`: add `self._session_agents`, an in-memory
  `sub_session_id -> raw agent_name` cache.
- `_spawn_new_session`: record `{sub_session_id: agent_name}` in the
  cache right after generating `sub_session_id` — this is the exact
  string later emitted in `delegate:agent_spawned`'s `"agent"` field,
  so spawned/resumed/completed events pair correctly under one
  identity for downstream counters.
- New `_resolve_agent_for_session()`: resolves agent identity for a
  `sub_session_id`. Priority:
  1. The spawn-time cache (exact match to `agent_spawned`'s value).
  2. Fallback: the session_id's sanitized suffix, per
     `amplifier_foundation.tracing.generate_sub_session_id`'s
     documented `{parent_span}-{child_span}_{sanitized_agent_name}`
     format — deterministic, not a guess, but lossy (lowercased,
     punctuation collapsed to hyphens). Used when the cache misses,
     e.g. a resume issued from a different parent session/process
     than the one that spawned the sub-session.
- `_resume_existing_session`: resolve `agent_name` up front (before
  the `try` block, so it stays bound in every `except` clause —
  pyright flagged this when I first computed it inside `try`) and
  include it in `delegate:agent_resumed`, `delegate:agent_completed`,
  `delegate:agent_cancelled`, and `delegate:error` payloads on this
  path.

Confined to `modules/tool-delegate/` — does not touch
`amplifier_foundation/spawn_utils.py` (reserved for another
in-flight PR).

## Tests

`modules/tool-delegate/tests/test_delegate_event_enrichment.py` —
new `TestResumeAgentIdentity`:
- `test_resumed_and_completed_events_carry_agent_from_spawn_cache`:
  full spawn → resume flow through `DelegateTool.execute()`, asserts
  `delegate:agent_resumed` and `delegate:agent_completed` both carry
  the real agent identity (not null/missing). **Fails with
  `KeyError: 'agent'` on unpatched main.**
- `test_resume_falls_back_to_session_id_suffix_when_cache_misses`:
  covers the cold-cache fallback path (session_id suffix parsing).
  Also fails on unpatched main.

## Test results

- `modules/tool-delegate/tests`: **62 passed** (60 pre-existing + 2 new).
- Repo suite (`uv run pytest tests/ -q --tb=short`): **1631 passed, 1 skipped**.
- `python_check` (ruff lint/format, pyright, stub check): clean —
  0 new warnings vs. unpatched main (verified via diff), 0 errors.

## Investigation: context-intelligence ingestion (secondary, no PR)

Per the request, I investigated whether the context-intelligence
server's session-completeness classification (`IncompleteSession` /
"has_terminal: false") mislabels resumed sub-sessions as a *local,
small* classification bug. It does not look like one:

- The classifier (`context_intelligence_server/handlers/data_layer_2/session.py`,
  `SessionHandler._handle_end` + `SessionLabelStateMachine.classify`)
  only marks `IncompleteSession` when `session:end` arrives and the
  node has **no prior type label** (`RootSession`/`SubSession`/`ForkedSession`)
  — i.e. its `session:start`/`session:fork` was never seen *in that
  same graph scope*.
- `session:resume` is **intentionally** excluded from
  `SessionHandler.handled_events` and routed to `DefaultHandler`
  instead — pinned by `tests/handlers/data_layer_2/test_session.py::TestSessionResumeNotClaimed`
  (`assert "session:resume" not in SessionHandler.handled_events`).
  A prior investigation (commit `8acdb3e`, "no bug in the label state
  machine") already audited a similarly-shaped anomaly and left this
  exclusion in place deliberately.
- I reproduced both shapes directly against the handlers: when
  `session:start` **is** seen before a `session:resume` + `session:end`
  sequence, the node correctly stays `SubSession` throughout (no
  `IncompleteSession`). Only when `session:resume` + `session:end`
  arrive with **no prior start/fork ever recorded** does the node get
  `IncompleteSession` — which is the classifier working exactly as
  documented, not a rule bug.

So the false positive isn't in the local classification rule; it
requires establishing *why* the resumed sub-session's original
`session:start`/`session:fork` never reached the same graph scope
before its `session:end` was processed — e.g. cross-run/cross-workspace
ingestion boundaries in the eval harness, or a gap in how a resumed
child session's own lifecycle hooks are wired in `amplifier-app-cli`'s
resume implementation (outside this repo). That's not a small,
confidently-scoped fix I can safely make without tracing production
event delivery across those boundaries, so I'm reporting it rather
than opening a second PR. Pointers for whoever picks this up:

- `context_intelligence_server/handlers/data_layer_2/session.py:118-136`
  (`SessionLabelStateMachine.classify`, `"end"` branch)
- `context_intelligence_server/handlers/data_layer_2/session.py:139-152`
  (`SessionHandler.handled_events` — deliberately excludes `session:resume`)
- `tests/handlers/data_layer_2/test_session.py:517-524`
  (`TestSessionResumeNotClaimed` — the pinned exclusion)

## Not included

- `amplifier-context-intelligence` PR — investigated, not opened (see above).

### DIFF
--- a/modules/tool-delegate/amplifier_module_tool_delegate/__init__.py
+++ b/modules/tool-delegate/amplifier_module_tool_delegate/__init__.py
@@ -133,6 +133,27 @@ class DelegateTool:
         # Build feature registry for dynamic description composition
         self._feature_registry = self._build_feature_registry()
 
+        # Maps sub_session_id -> the raw (un-sanitized) agent_name recorded at
+        # spawn time. This is the authoritative source for agent identity on
+        # resume: it is the exact same string emitted in delegate:agent_spawned,
+        # so counters that pair spawned/resumed/completed events by "agent"
+        # stay correct. Populated in _spawn_new_session(); consulted in
+        # _resume_existing_session() via _resolve_agent_for_session().
+        #
+        # Scope: this cache lives on the DelegateTool instance, which is
+        # constructed once per mounted session (see mount()) and persists for
+        # the lifetime of the parent session/coordinator. It does NOT survive
+        # across process restarts or a resume issued from a *different*
+        # parent session than the one that spawned the sub-session -- that
+        # cold-cache case falls back to parsing the agent name out of the
+        # session_id suffix (see _resolve_agent_for_session), which is lossy
+        # (sanitized: lowercased, non-alphanumeric chars collapsed to hyphens)
+        # but always available since
+        # amplifier_foundation.tracing.generate_sub_session_id guarantees the
+        # "{parent_span}-{child_span}_{sanitized_agent_name}" shape for every
+        # sub-session id this tool creates.
+        self._session_agents: dict[str, str] = {}
+
     def _build_feature_registry(self) -> list[dict[str, Any]]:
         """Build registry of features with their descriptions.
 
@@ -995,6 +1016,12 @@ Agent usage notes:
             parent_session_id=parent_session_id,
         )
 
+        # Record the raw agent_name for this sub-session so a later resume
+        # (via _resume_existing_session) can recover the exact identity used
+        # here, instead of re-deriving a lossy, sanitized approximation from
+        # the session_id suffix. See _resolve_agent_for_session().
+        self._session_agents[sub_session_id] = agent_name
+
         # Resolve agents from coordinator config if not provided directly
         if agents is None:
             try:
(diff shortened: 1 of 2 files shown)


bkrabach:
**Maintainer admin-merge basis (documented per team policy):**

- **Author:** this PR was authored by our own automated account. GitHub structurally forbids self-approval, so no self-review can satisfy branch protection here — an external human reviewer is not in the loop for this change.
- **Test evidence:** carries agent identity through resume completions via a spawn-time cache plus session-id-suffix fallback, and emits `agent_resumed`. 2 new tests added, both fail-before/pass-after regression tests. Full repo suite: 1631 passed.
- **CI status at merge time:** all 6 matrix jobs (ubuntu/windows × Python 3.11/3.12/3.13) + license/cla check green. See status checks on this PR.
- **Action:** merging via maintainer admin override (`gh pr merge --admin`) at the explicit direction of the repository owner, following the same sanctioned pattern previously used for gemini#39 (self-authored, self-approval-impossible, explicit maintainer direction).

Squash-merging now.

PR #317, opened 2026-08-27
fix(agents): idempotency discipline for git-ops (context-truncation thrash)

bkrabach:
## Summary

- Adds a compact **Idempotency Discipline** checklist to `agents/git-ops.md`, positioned right after the existing "Git Safety Protocol" section (near the top of operational guidance, before the command references) so it survives context truncation.
- Addresses work item `openai_improvement-bq0`: 5 production eval runs (fast-tier model, `enable_long_context=false`) showed git-ops re-running `git clone --branch main --single-branch <same URL>` up to 8x and `git ls-remote --symref <same URL>` up to 7x within a single session (one run: 30 bash calls, only 14 unique) — the signature of context truncation losing the memory that setup work already completed. Sessions ran 809–1168s despite the "fast" label.
- Prompt/agent-definition change only — no Python touched.

## Added text (exact)

```
## Idempotency Discipline

Disk and repo state are the source of truth — not your memory of what you already ran.

**Before ANY clone:** check whether the target directory already exists with the right remote (`git -C <dir> remote get-url origin`). If it matches, `git -C <dir> fetch` instead of re-cloning. Never clone the same URL twice in one session.

**Remote-discovery commands run at most once per remote per session.** `git ls-remote`, default-branch lookups, and similar discovery calls are one-shot — write the result into your working notes immediately and reuse it. Don't re-derive it.

**About to repeat a command you believe you already ran? Stop.** Check the filesystem/repo state first. Repeated identical commands are the signature of lost context, not a reason to run them again.
```

Inserted between the existing "Git Safety Protocol" and "Common Git Commands" sections (`agents/git-ops.md:124-133` on this branch).

## Test plan

- [x] N/A — prompt/agent-definition-only change (Markdown frontmatter + prose), no code paths to unit test.
- [x] Diff reviewed for minimality: 10 lines added, 0 removed, no restructuring of surrounding sections.
- [ ] Recommend validating with a DTU/eval re-run against the same fast-tier/`enable_long_context=false` matrix entry that produced the original evidence, to confirm the repeat-clone/repeat-ls-remote pattern no longer recurs.

## Recommendation for routing-matrix maintainers (informational — no code change here)

[...]

PR #320, opened 2026-08-27
feat(delegate): plan-then-dispatch guidance — batch independent delegations in one turn

bkrabach:
## Summary

Stage 1 ("Option A", guidance-only) of the approved `pg1-parallel-delegation` spec. Text-only guidance edits — no code, schema, or behavior changes.

## Measured evidence

A measured eval showed root sessions delegating in **3–4 sequential waves**, idle **91–94% of wall time**, with **~57% of wall structurally recoverable** by batching independent delegations into a single turn. One eval run's wave 2 was `[explorer, explorer, git-ops]` — none of which consumed wave 1's output. The root had all the information needed to batch those into wave 1 and did not.

## Why guidance, not machinery

The parallel-dispatch machinery already works end-to-end and needed no changes:

- The orchestrator mints one `parallel_group_id` per assistant turn and runs every tool call from that turn concurrently via `asyncio.gather` (`amplifier_module_loop_streaming`).
- The OpenAI provider already sets `parallel_tool_calls=True` by default; Anthropic never disables it.
- Concurrent delegates don't race on dispatch context (task-keyed, not session-keyed).

**There is no mechanical serialization to remove.** A root that emits three `delegate` calls in one turn already gets three concurrent sub-sessions. The measured idle time is the machinery faithfully executing a *one-delegation-per-turn plan* — a planning-policy defect, not an execution-machinery defect. Per `KERNEL_PHILOSOPHY.md` §1, policy belongs at the edges (instructions, tool descriptions), not in new kernel-adjacent code.

Every existing mention of parallelism in this repo argued **quality** ("different agents bring different tools") — none argued **wall-clock**, and none told the model to enumerate its full delegation set before dispatching. This PR fixes that gap directly at its source.

A companion approach (a `delegations[]` array with explicit dependency edges) was evaluated and rejected in the spec: it asks the model for exactly the judgment whose absence is the defect, re-implements `asyncio.gather` one layer below where it already exists correctly, and its marginal wall-clock gain over a correctly-batched guidance fix is under 1%.

## Changes (3 verbatim text replacements, +175 tokens/request budgeted)

- `modules/tool-delegate/amplifier_module_tool_delegate/__init__.py` — replaced the delegate tool's "Agent usage notes" block (re-rendered into every request's tool spec, last thing read before a `delegate` call) with explicit batching guidance and an independence heuristic.
- `context/agents/multi-agent-patterns.md` — replaced "## Parallel Agent Dispatch" (framed only as "richer results") with "## Plan the Whole Set, Then Dispatch", adding the wall-clock framing and an explicit independence test.
- `context/agents/delegation-instructions.md` — replaced "## Scaling with Multiple Instances" (same-agent only) with "## Wave Discipline: Batch Everything Independent", covering both same-agent and cross-agent batching.

Each edit *replaces* its section rather than appending, per `ISSUE_HANDLING.md`'s guidance on not growing files without cause. No file grows a section count.

## Verification

- `uv run pytest tests/ -q --tb=short` (the exact CI command): **1634 passed, 1 skipped**
- `modules/tool-delegate/tests/` (62 tests, editable-installed locally to resolve the package): **62 passed** — the existing description/schema tests are unaffected by the wording change
- `python_check` on the touched `.py` file: clean except 8 pre-existing warnings elsewhere in the file (confirmed present before this change via `git stash`), none introduced by this edit
- Confirmed no `return` + `{` substring was introduced — `amplifier_foundation.bundle_docs.tool_schema._extract_input_schema` still successfully `ast.literal_eval`-extracts the tool's static input schema after the edit
- Grepped `tests/` and `modules/*/tests/` repo-wide for the replaced phrases (`"Launch multiple agents concurrently"`, `"Parallel Agent Dispatch"`, `"Scaling with Multiple Instances"`, etc.) — no test asserts the old wording, so no test updates were needed

## Deferred (not in this PR)

**Stage 2** (`hooks-delegation-batching`, a nudge hook that reinforces this guidance at the actual decision point — after wave N's results land, before wave N+1 is planned) is fully specified in the spec but **not implemented here**. Per the spec's rollout plan, it ships only if Stage 1's DTU measurement (median wall-time reduction) misses the ≥40% target. If Stage 1 alone hits target, Stage 2 never ships.

**B′** (a flat `delegations[]` array, no dependency edges) is kept on the shelf as a contingency, gated on a specific signature (a backend whose tool-call parser structurally emits at most one call per turn) that has not been observed. Not scheduled.

## Post-merge note

After merge, `amplifier reset --remove cache -y` is required before the new guidance is loaded by a running Amplifier process — foundation context files are read from the bundle cache, and the guidance is prose read at composition time, not code.

### DIFF
--- a/context/agents/delegation-instructions.md
+++ b/context/agents/delegation-instructions.md
@@ -276,21 +276,34 @@ Use session resumption when:
 
 ---
 
-## Scaling with Multiple Instances
+## Wave Discipline: Batch Everything Independent
 
-For large codebases or complex investigations, dispatch MULTIPLE instances of the same agent with different scopes:
+**Every delegate call in one turn runs concurrently. Every delegate call in a separate turn
+runs sequentially, and you block on the previous one first.**
+
+Plan your full delegation set before dispatching any of it. Emit every independently
+resolvable delegation in a single turn — different agents, same agent with different
+scopes, or both.
 
 ```python
-# Parallel dispatch - independent surveys
+# One turn, three concurrent agents - different specialists
+delegate(agent="foundation:explorer", instruction="Survey auth/", context_depth="none")
+delegate(agent="python-dev:code-intel", instruction="Trace authenticate() callers")
+delegate(agent="foundation:git-ops", instruction="Summarize recent auth/ commits")
+
+# One turn, three concurrent instances - same agent, split scope
 delegate(agent="foundation:explorer", instruction="Survey auth/", context_depth="none")
 delegate(agent="foundation:explorer", instruction="Survey api/", context_depth="none")
 delegate(agent="foundation:explorer", instruction="Survey models/", context_depth="none")
 ```
 
-**When to scale:**
-- Large codebase with distinct areas
-- Multiple independent questions to answer
-- Time-sensitive investigations where parallelism helps
+**Only a delegation that consumes another delegation's output belongs in a later turn.**
+Everything else goes now. Before you dispatch a second wave, ask whether it could have
+gone out with the first — if it could have, batch what remains rather than repeating the
+mistake.
+
+**When to scale out:** large codebase with distinct areas; multiple independent questions;
+any investigation where you would otherwise wait on one agent before starting the next.
 
 ---
 
--- a/context/agents/multi-agent-patterns.md
+++ b/context/agents/multi-agent-patterns.md
@@ -4,22 +4,39 @@ This context provides patterns for orchestrating multiple agents effectively.
 
 ---
 
-## Parallel Agent Dispatch
+## Plan the Whole Set, Then Dispatch
 
-**CRITICAL**: For non-trivial investigations or tasks, use MULTIPLE agents to get richer results. Different agents have different tools, perspectives, and context that complement each other.
+**Delegations emitted in ONE turn run concurrently. Delegations split across turns run
+sequentially — and you sit idle through each one.**
 
-When investigating or analyzing, dispatch multiple agents IN PARALLEL in a single message:
+Before your first delegate call, write down every delegation the task needs. Then emit,
+in a single turn, every one of them that does not consume another delegation's output.
 
 ```python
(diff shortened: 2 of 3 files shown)


bkrabach:
**Merge basis (documented admin-merge — precedent: #317/#318/#319, gemini#39, merged same-day under this pattern)**

- **Test suite:** full suite green — 1634 passed.
- **DTU A/B measurement:** Live A/B in a Digital Twin Universe measured the tool-description edit directly:
  - First-delegation-wave width rose **1 → 3**, with **zero dependency-breaking batches** observed.
  - Token cost held within **±6%** of baseline.
  - Wave-count and wall-clock targets were **not** met in this harness — the harness loads only 1 of the 3 edited files, which understates the effect of the full guidance change. A clean re-measure (all 3 files loaded) is in flight.
- **Net assessment:** directionally positive (more parallel delegation, no correctness regressions, negligible token overhead), zero-risk guidance-only change (no behavioral/code-path changes, docs/prompt-guidance only), with an outstanding clean re-measure to fully confirm the wave/wall targets.
- **Review gate:** PR is self-authored; GitHub blocks self-approval, so `reviewDecision` is stuck at `REVIEW_REQUIRED` with all required checks green (`mergeStateStatus: BLOCKED` on review only, not checks).
- **Action:** merging via `--admin` at explicit user direction, consistent with the established maintainer admin-merge pattern used for #317, #318, #319, and gemini#39 earlier today.

bkrabach:
CORRECTION/OUTCOME: a clean same-foundation-commit 5v5 DTU A/B measured this guidance INERT — treatment waves median 4 vs control 3, wall/token deltas within noise; the earlier apparent front-loading win was a foundation-version confound. The tool-description block is being removed in #328 (~175 tokens/request with no measured benefit). The two context-file sections remain (net-zero replacements). Lesson recorded: prose guidance does not change delegation behavior on these models; mechanical levers do.

PR #322, opened 2026-08-27
fix(delegate): structured event + optional strict mode for unresolved model_role

bkrabach:
## Summary

`resolver.resolve(raw_model_role)` returning no candidates was silently
absorbed: `tool-delegate` logged a single WARNING and proceeded with
`provider_preferences=None`, so the delegation quietly landed on the
session's default model with no signal an operator could act on — including
after `list_models` retry-exhaustion during a persistent provider outage
(the resolver has nothing to match against once the provider stops
answering).

This is the **third member of the silent-substitution family** found in
this codebase, after:
- #318 — sub-agent provider override losing a priority-0 tie-break (fixed by
  strictly demoting other instances instead of merely tying)
- the `list_models` retry work — transient provider failures previously
  looked identical to "provider has zero models" from the caller's side

All three share the same shape: a resolution step fails or comes back
empty, and the system falls back to *some* default without telling anyone
it did. This fix is cross-provider by construction — nothing here is
Anthropic/OpenAI/Gemini-specific; any `model_role_resolver` capability hits
the same code path.

## Changes

`modules/tool-delegate/amplifier_module_tool_delegate/__init__.py`

1. **Structured event, always emitted.** On the no-candidates path, in
   addition to the existing WARNING log, we now emit
   `delegate:model_role_unresolved` via the same `hooks.emit()` mechanism
   used for `delegate:agent_spawned` (registered in the module's
   `observability.events` contribution list at mount time):

   ```python
   {
       "model_role": raw_model_role,
       "agent": agent_name,
       "resolver": resolver_name,   # class/name of the resolver instance
       "fallback_behavior": "session_default",
   }
   ```

   This is unconditional — emitted regardless of `strict_model_role` — so
   the substitution is observable even for callers who haven't opted into
   fail-loud mode.

2. **New `settings.strict_model_role` (bool, default `False`).** When
   `True`, the no-candidates path raises `ModelRoleUnresolvedError` (a new
   `RuntimeError` subclass) naming both the role and the resolver, instead
   of falling back:

   ```
   model_role 'coding' resolved to no candidates against installed
   providers (resolver=test-matrix). strict_model_role is enabled, so this
   delegation is refused instead of silently substituting the session
   default model.
   ```

   Default `False` preserves the exact current behavior for every existing
   caller that relies on the quiet fallback — this is additive, not a
   breaking change.

## Sibling path investigated (caller-passed `provider_preferences` glob)

Per the task, I traced the other place a "resolves to nothing" silent
substitution could occur: a caller passing `provider_preferences` directly
with a glob `model` (e.g. `"gpt-9-*"`) that matches no installed model.
That path lives in `amplifier_foundation/spawn_utils.py`
(`resolve_model_pattern` / `apply_provider_preferences_with_resolution`),
not in `tool-delegate` itself — `tool-delegate` just parses and forwards
`ProviderPreference` objects; resolution happens deeper in the spawn
pipeline (`amplifier_foundation/bundle/_prepared.py`).

Verdict: **for a single preference, this already fails loud** — a
non-matching glob returns an explicit `resolved_model=None` sentinel (never
substitutes the raw/broken pattern) and the caller deliberately advances to
the *next* preference in the ordered fallback list (correct semantics for
an ordered list, and already carefully commented as intentional). Per the
task's instruction ("if it already fails loud, note it and leave it"), I
left this untouched.

One narrower residual gap, noted here rather than folded into this PR: if
**every** preference in the caller's list fails to resolve (or none of
their providers are present in the mount plan), the function falls back to
the *unmodified* mount plan with a warning-only log — the same
class of silent fallback as the bug this PR fixes, but in a much
lower-level, broadly-shared utility (used by every `session.spawn` /
`prepared.spawn()` caller, not just delegate) with no existing `hooks`/
`agent` parameter threading. Extending the same event+strict-mode
treatment there is a reasonable follow-up but is a separate, larger-blast-
radius change than this module-scoped fix warrants on its own.

## Tests

`modules/tool-delegate/tests/test_delegate_model_role.py` —
new `TestModelRoleUnresolvedEvent` class:

- `test_default_mode_emits_event_and_preserves_fallback` — event emitted
  with the documented payload, delegation still succeeds,
  `provider_preferences` stays `None` (fallback preserved). **Fails on
  unpatched main**, which never emits any event on this path.
- `test_strict_mode_raises_instead_of_falling_back` — raises
  `ModelRoleUnresolvedError` naming both the role (`coding`) and the
  resolver (`test-matrix`); `spawn_fn` never called; event still emitted.
- `test_normal_resolution_emits_no_unresolved_event` — resolver returning
  candidates emits no `delegate:model_role_unresolved` event.
- `test_strict_mode_defaults_to_false` — confirms the default.

Also updated `test_delegate_contribution_channel.py`'s exact-event-count
assertion (5 → 6) since the module now legitimately contributes one more
observability event.

### Results

```
# Full repo suite (uv run pytest tests/ -q --tb=short)
1634 passed, 1 skipped   (unchanged from baseline on this main)

# tool-delegate module suite
66 passed   (was 62 before this change; +4 new tests)
```

## Not merging

Per instructions, opening for review only — not merging.

### DIFF
--- a/modules/tool-delegate/amplifier_module_tool_delegate/__init__.py
+++ b/modules/tool-delegate/amplifier_module_tool_delegate/__init__.py
@@ -20,6 +20,12 @@ Config Options:
 - settings.exclude_tools: Tools spawned agents should NOT inherit (default: ["tool-delegate"])
 - settings.exclude_hooks: Hooks spawned agents should NOT inherit (default: [])
 - settings.timeout: Maximum total execution time for child session in seconds (default: None/disabled)
+- settings.strict_model_role: When True, a model_role that resolves to no
+  candidates raises ModelRoleUnresolvedError instead of silently falling
+  back to the session default model (default: False). Regardless of this
+  setting, the no-candidates case always emits a
+  "delegate:model_role_unresolved" event so the silent substitution is
+  observable.
 
 Tool Parameters:
 - agent: Agent to delegate to (e.g., 'foundation:explorer', 'self')
@@ -43,6 +49,18 @@ from amplifier_foundation import ProviderPreference
 from amplifier_foundation.tracing import generate_sub_session_id
 
 logger = logging.getLogger(__name__)
+
+
+class ModelRoleUnresolvedError(RuntimeError):
+    """Raised when ``model_role`` resolves to no candidates under strict mode.
+
+    Only raised when ``settings.strict_model_role`` is ``True``. Signals
+    that the requested ``model_role`` could not be resolved to any
+    provider/model candidate against the installed providers, and the
+    caller has opted out of the default silent-fallback-to-session-default
+    behavior. See ``delegate:model_role_unresolved`` for the always-emitted
+    observability event that accompanies both the strict and default paths.
+    """
 
 
 async def mount(coordinator: ModuleCoordinator, config: dict[str, Any] | None = None):
@@ -70,6 +88,7 @@ async def mount(coordinator: ModuleCoordinator, config: dict[str, Any] | None = 
             "delegate:agent_completed",
             "delegate:agent_cancelled",
             "delegate:error",
+            "delegate:model_role_unresolved",
         ],
     )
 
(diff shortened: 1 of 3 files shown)


bkrabach:
Merging per this week's silent-substitution remediation family — third member after #318 (tiebreak) and the list_models retry fixes. Adds an always-on `delegate:model_role_unresolved` structured event plus an optional `strict_model_role` flag (default off, behavior-preserving). 66 module tests + 1634 repo tests green; full CI matrix (ubuntu/windows × 3.11/3.12/3.13) passing.

Self-authored PR — self-approval is impossible via GitHub's review model, so this is an admin merge at user direction, consistent with this week's precedent for the rest of the family.

PR #298, opened 2026-08-07
fix(delegate): bound delegated sessions by default

ramparte:
## Summary

Real-world runaway delegated sessions have reached **17h34m / 8,036 calls** and **3h49m / 992 calls**. This adds a delegate-owned circuit breaker without exposing or depending on any private session data.

- Default `settings.timeout` to **1800 seconds** for both spawn and resume; an explicit `null` remains the opt-out for unbounded delegation.
- Validate timeout configuration eagerly as a positive, finite, non-boolean number that the event loop can represent.
- Enforce hard parent-release semantics: at the deadline, cancel the child and return the timeout result immediately rather than waiting for cancellation handling or slow persistence/cleanup to finish.
- Retain detached child tasks strongly until completion, consume their terminal results, and release them without "task destroyed" or unobserved-exception leakage.
- Report recovery state honestly as `status: timed_out`, `resumable: false`, and `resume_status: pending_child_cleanup`; the response never claims that interrupted-session persistence has completed.
- Emit `delegate:error` with `error_type: delegate_timeout`, not `delegate:agent_completed`, while preserving ordinary external-cancellation semantics.
- Return the child session ID and agent identity when available so recovery can be correlated after cleanup completes.
- Document the default and explicit-null opt-out, add focused validation/deadline/detached-task/error-semantics coverage, and include module tests in default pytest collection.
- Ignore `.next/` working-session artifacts.

## Coordinated prerequisite

This rollout depends on microsoft/amplifier-app-cli#260, which persists interrupted child sessions and guarantees shielded cleanup after the Foundation layer releases the parent at its deadline. The pending-cleanup response is deliberately conservative until that app-layer work completes.

## Verification

- 30 focused timeout/circuit-breaker tests passed.
- 90 `tool-delegate` module tests passed.
- 1,639 full-suite tests passed with one pre-existing warning.
- Ruff check clean.
- Ruff format check clean.
- Pyright clean.

PR #325, opened 2026-08-27
feat(delegate): flag-gated per-leg call budget via orchestrator max_iterations (Layer 1)

bkrabach:
## Summary

Layer 1 of the "Layered Bounding for Delegated Sessions" spec (298-replacement — replaces the wall-clock-only default in #298, which is being revised separately rather than merged as-is). Gives `tool-delegate` the ability to inject a per-session-leg LLM-call budget into a child session, using the delegate's existing `orchestrator_config` channel — zero new kernel surface, zero app-cli changes.

Enforcement lives in the child's own orchestrator loop (`max_iterations`), not in this module — this module only resolves *what value* to inject and reports what came back. See companion PR microsoft/amplifier-module-loop-streaming#43 for the orchestrator-side work this depends on for the budget to actually do anything.

**Ships DARK.** `settings.max_llm_calls` defaults to `None`. With no config change, `orchestrator_config` is byte-for-byte what it was before this PR — nothing is injected, nothing changes.

## Precedence chain

| Rank | Source | Status |
|---|---|---|
| 1 | Per-call tool input (`max_llm_calls`) | ✅ implemented |
| 2 | Per-agent frontmatter (`agents[name]["budget"]["max_llm_calls"]`) | ❌ **not implemented — see below** |
| 3 | `settings.max_llm_calls` (default `None`) | ✅ implemented |
| 4 | Inherited parent `orchestrator_config.max_iterations` | ✅ implemented (pre-existing path, untouched when no budget applies) |

An explicit `0` at rank 1 means "no Layer 1 budget for this call" (the wall-clock `settings.timeout` backstop still applies) and correctly overrides rank 3's default — this required care: a naive "0 → None" collapse done too early loses the ability to distinguish "explicitly opted out" from "not specified at all," which would have silently ignored a caller's opt-out. Covered by `TestOptOut` in the new test file.

## Per-agent frontmatter override did NOT make the cut — verified, not assumed

The spec flagged this as an open item requiring verification before shipping: does a top-level `budget:` block in an agent `.md` file's frontmatter actually reach `coordinator.config["agents"][name]["budget"]`?

**Verified empirically** (wrote and ran a test against the real loader, not just read the source): it does **not** round-trip. `amplifier_foundation.bundle._dataclass._load_agent_file_metadata` only forwards a fixed allowlist of top-level frontmatter keys (`tools`, `providers`, `hooks`, `session`, `provider_preferences`, `model_role`, `agents`) — `budget` is silently dropped. See `test_agent_frontmatter_budget_key_is_dropped` in the new test file (uses `model_role`, an allowlisted key, as a control to prove this isn't a wholesale frontmatter-loading failure).

Per the owner's decision, rank 2 is **left out of this PR** rather than faked. Ranks 1/3/4 ship and are fully tested. Closing this gap requires a change to `amplifier_foundation.bundle._dataclass._load_agent_file_metadata` (add `budget` to the allowlist) — filed as a follow-up rather than bundled here, since it's a foundation-loader change, not a tool-delegate change. See the module README's "Known gaps" section.

## Also added

- **Eager validation** (`_check_call_budget_type` / `_validate_call_budget`): rejects `bool`, non-`int`, and negative values at the point supplied — module construction for the settings default, `execute()` for the per-call override — never silently coerced, never deferred to spawn time.
- **Negotiated-feature warning** (spec §4.4): if a budget was requested but the child's orchestrator reports no `llm_call_budget` telemetry (e.g. a third-party orchestrator without `max_iterations` support), logs a warning naming the agent and sets `metadata.budget_enforced = false` on the returned `ToolResult` — silence is exactly the failure mode this design exists to eliminate.
- `max_llm_calls` entry in the tool's input schema (kept as a pure literal for the static token-cost estimator, per this file's existing convention).

## Files changed

- `modules/tool-delegate/amplifier_module_tool_delegate/__init__.py`:
  - `_check_call_budget_type` / `_validate_call_budget` (module-level validation helpers)
  - `__init__`: `settings.max_llm_calls` / `budget_warn_ratio`
  - `execute()`: per-call `max_llm_calls` parsing + validation
  - `_resolve_call_budget` (new method): 2-rank precedence (1 and 3; rank 4 falls through untouched)
  - `_spawn_new_session`: `orchestrator_config` build now copies (never mutates) the parent's config before injecting the budget; negotiated-feature warning + `budget_enforced` metadata
  - `_static_input_schema()`: `max_llm_calls` parameter
- `modules/tool-delegate/README.md`: "Layer 1 call budget" section + "Known gaps"
- `modules/tool-delegate/tests/test_delegate_call_budget.py` (new): 14 tests covering the precedence chain, opt-out, validation, status/metadata passthrough, the negotiated-feature warning, resume behavior, and the frontmatter verification

## Testing

- New tests: `uv run pytest modules/tool-delegate/tests/test_delegate_call_budget.py -q` → **14 passed**
- Full tool-delegate module suite: `uv run pytest modules/tool-delegate/tests/ -q` → **80 passed** (zero regressions)
- Full foundation repo suite: `uv run pytest tests/ -q` → **1634 passed, 1 skipped** — identical to the pre-change baseline
- `python_check` (ruff lint/format, pyright): **no new issues** vs. baseline (all remaining warnings/errors are pre-existing, same lines shifted, confirmed via before/after diff)

## Relationship to other PRs

- Companion PR: microsoft/amplifier-module-loop-streaming#43 (orchestrator-side enforcement — required for this PR's config to have any effect)
- #298 is being revised separately (not closed) to raise its wall-clock default 1800s → 14400s and reframe as the Layer 3 backstop that sits *behind* this budget

**Do not merge** — up for review alongside the companion PR and the #298 revision.

### DIFF
--- a/modules/tool-delegate/README.md
+++ b/modules/tool-delegate/README.md
@@ -45,6 +45,7 @@ Agent's explicit tool declarations are always honored, even when parent excludes
 | `context_turns` | integer | 5 | Number of turns when context_depth is 'recent' |
 | `context_scope` | enum | "conversation" | Which content: conversation, agents, full |
 | `provider_preferences` | array | - | Ordered provider/model preferences |
+| `max_llm_calls` | integer | - | Override the Layer 1 LLM-call budget for this delegation (per session leg). `0` disables the budget for this call. Only takes effect when `settings.max_llm_calls` is configured -- see "Layer 1 call budget" below. |
 
 ## Configuration
 
@@ -70,6 +71,16 @@ modules:
         - delegate  # Default: spawned agents can't further delegate
       exclude_hooks: []
       timeout: 300
+      max_llm_calls: null      # Layer 1 call budget (spec: 298-replacement).
+                                # None/unset (default): ships dark -- no
+                                # budget is injected into any child session;
+                                # today's behavior is unchanged. Set to a
+                                # positive integer to enforce a per-leg
+                                # LLM-call budget (see below). 0 is
+                                # equivalent to null (explicit no-budget).
+      budget_warn_ratio: 0.8    # Fraction of max_llm_calls at which a
+                                # one-shot "start converging" warning fires.
+                                # Only meaningful when max_llm_calls is set.
 ```
 
 ### Structured Return Contract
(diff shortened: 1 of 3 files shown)


bkrabach:
Rebased `feat/delegate-call-budget` onto current `main` (which now includes #322, #323, and #324) to resolve the CONFLICTING/DIRTY state reported after #324 merged (squash `1439cf5`) ahead of this PR.

**Conflict resolution — both additive features kept in full, nothing dropped:**
- `modules/tool-delegate/amplifier_module_tool_delegate/__init__.py`:
  - Config-parsing / module-header region: kept #324's `RETURN_CONTRACT_SCHEMA` / `RETURN_CONTRACT_INSTRUCTION` / `_return_contract_event_fields` / `_JSON_FENCE_PATTERN` block, followed by this PR's `_check_call_budget_type` / `_validate_call_budget` helpers, ahead of `mount()`.
  - Return-path region: merged #324's annotated-output comment with this PR's `output_metadata` construction (`budget_enforced` merged into the metadata bag), then combined both additive output keys — `"metadata": output_metadata` (this PR) and `"contract": contract` (#324) — into the same `ToolResult`.
- `modules/tool-delegate/README.md`: kept both new sections — #324's "Structured Return Contract" followed by this PR's "Layer 1 call budget" — no content removed from either.

**Testing (post-rebase, both features exercised together):**
- `modules/tool-delegate/tests/` (module suite): **120 passed** — includes #324's return-contract tests (40, across 4 files) and this PR's call-budget tests (14, `test_delegate_call_budget.py`) running side by side with no interaction regressions; both parse cleanly from the same `config`/`settings` dict.
- Full repo suite `uv run pytest tests/`: **1634 passed, 1 skipped**.
- `python_check` (ruff lint/format, pyright) on the touched module file: 0 errors; only pre-existing warnings inherited from `main` (verified against `main`'s copy of the file) remain — a trivial pre-existing `ruff format` line-length nit from this PR's own diff was fixed in the process.

Ships dark: `settings.max_llm_calls` defaults to `None` — no behavior change for existing users until an operator opts in.

Self-authored / merging with admin at user direction, per this week's precedent for author-approved same-repo branches.

PR #336, opened 2026-08-31
feat(tool-delegate): optional max_spawns_per_session cap on new sub-agent spawns

bkrabach:
## Summary

Adds an optional, defensively-parsed `settings.max_spawns_per_session` cap to
`DelegateTool` (`modules/tool-delegate`): once a parent session has spawned
that many **new** sub-agent sessions, further new-spawn requests get a normal
(non-exception) declined `ToolResult` naming the cap and pointing at the
caller's own tools, while session **resumes** remain completely unmetered.
Default (absent/`None`) is **unlimited — byte-identical to today's behavior**.

## Why

Models with a strong intrinsic delegation habit (gpt-5.6-sol-class) over-delegate
regardless of tool-description wording:

- On an identical background task, **claude-opus** solved it single-session
  with **zero** delegation, while **sol** spawned **6–11** sub-agents.
- Two prior treatments targeted only the tool description's *framing*:
  - T-dg: a relaxed decision-rule rewrite → median **7** spawns
  - T-strip: full advocacy+example strip (description 14,562 → 8,060 chars) → median **9** spawns
  - Both against a control median of **6**, with quality held at **18/18** in both.
- **Rewording the description did not move the needle.** That is the evidence
  this is model-intrinsic, not a fixable-by-better-prose problem — the
  pre-registered fallback for a model-intrinsic behavior is a mechanical
  constraint, not a fourth wording attempt.
- OpenAI's own hosted multi-agent documentation recommends a max concurrency
  of 3, which is the reference value used throughout validation below (the
  actual value is fully operator-configured, not hardcoded in the patch).

## Validation (DTU A/B, dial-in wave, `max_spawns_per_session: 3`)

- Root new-spawns stayed **≤ 3** on every capped run (vs **11** and **5**
  uncapped) — and the model **voluntarily stopped at or below the cap**: the
  budget-reached decline path never actually fired across the wave.
- Quality fully held: single-turn **2/2** correct; multi-turn **2/2** at
  **100/100** grader score.
- Session count dropped **12 → 4**.
- **claude-opus unaffected**: 0 delegations with and without the cap — it
  never binds, consistent with T-strip's earlier finding that opus is
  largely insensitive to the tool description's specific content.

**Honest framing:** the cap is a *construction guarantee* (mechanically true
once configured), not a claimed change in model behavior — the model chose to
respect it early, which is a genuinely interesting secondary result but not
the point of the feature. Cost savings are **secondary and roughly a wash**
(capped runs did the remaining work in-root at similar cost); the real value
is bounded fan-out, fewer sessions, and cache-friendlier session trees —
structural discipline, not a cost play.

## What changed

`modules/tool-delegate/amplifier_module_tool_delegate/__init__.py` (+151/-0, 5 hunks):

- `_parse_spawn_budget(value)` — defensive parser. `None` → unlimited (default).
  `int` (not `bool`) ≥ 0 → itself. `str` → coerces via `int(value.strip())`
  when non-negative. Anything else (negative, `bool`, `float`, unparseable
  string, any other type) **warns and falls back to unlimited — never raises**.
  (Deliberately different from the sibling `max_llm_calls` budget, which
  raises at construction — a misconfigured spawn budget must never take down
  the whole delegate tool.)
- `DelegateTool.__init__`: parses the setting once; `self._new_spawns_used = 0`.
- `DelegateTool.execute()`: in the SPAWN branch only (resumes are routed away
  earlier via `session_id` and never reach this check) — at/over the cap,
  returns `ToolResult(success=False, error={"message": ..., "code":
  "SPAWN_BUDGET_EXCEEDED"})` and **spawns nothing** (`session.spawn` is never
  called on this path).
- `DelegateTool._spawn_new_session()`: counts **only on the success path**
  (never on the `_DelegateTimeoutExpired` or generic-exception paths, so an
  infra failure never costs the caller a slot) and appends a
  `"(new-session budget: X of Y remaining)"` footer to the response.
- `DelegateTool.description`: one factual sentence appended when a budget is
  configured; nothing added when unlimited.
- `0` is a valid, literal budget (blocks every new spawn) — **not** collapsed
  to "unlimited" the way `0` is for the sibling `max_llm_calls` budget, since
  the task spec only treats *absent/None* as unlimited for this key.
- Decline message is 32 words (asserted `<=60` by test), names the caller's
  own remaining tools (filesystem, search, bash), and explicitly says
  resuming an existing session is still allowed.

`modules/tool-delegate/tests/test_delegate_spawn_budget.py` (new, +324/-0):
18 tests / 6 classes — default-unlimited zero-behavior-change,
budget-enforcement (3 succeed w/ footers, 4th rejected + spawns nothing),
resume-never-counts, defensive parsing (string coercion, garbage → warn +
unlimited), garbage-value end-to-end, and description-line presence.

**Known, disclosed limitation:** the pre-spawn check and the post-spawn
counter increment are not atomic across concurrent invocations, so a burst of
simultaneous delegate calls (e.g. an orchestrator's own `asyncio.gather`) could
allow limited overshoot beyond the configured cap. This mirrors an existing,
already-accepted pattern in the same file (the self-delegation-depth check
reads a capability with no lock either) and is a mechanical backstop, not a
hard security boundary.

## Test evidence

Run against this exact branch, both before (`git stash -u`) and after
(`git stash pop`) on the same checkout — not assumed from the diff:

| Suite | Before | After | Delta |
|---|---|---|---|
| `modules/tool-delegate/tests` | 150 passed | **168 passed** | +18 (new), 0 regressions |
| Full repo (`uv run pytest`) | 1793 passed, 1 skipped | **1811 passed, 1 skipped** | +18 (new), 0 regressions |

`python_check` (ruff format / ruff lint / pyright) on both touched files:
10 pre-existing warnings on `__init__.py`, identical codes at shifted line
positions (confirmed via the same stash/pop comparison — this patch is a
pure insertion, `+151/-0`); 0 new lint/format issues. The new test file's
one `pyright reportMissingImports` (`amplifier_module_tool_delegate`) is the
same pre-existing false positive already present on the neighboring,
untouched `test_delegate_call_budget.py` (confirmed directly) — `pytest`'s
own `pythonpath` setting resolves it at runtime; the 168-pass run above is
the real signal.

## Config (opt-in, no default behavior change)

```yaml
modules:
  tool-delegate:
    settings:
      max_spawns_per_session: 3  # optional; absent/null = unlimited (today's behavior)
```

## Drift / rebase note

Built against `amplifier-foundation@3a43b29`. `main` has since moved to
`4d22a3a` ("fix: preserve git source subdirectories") — confirmed that commit
does not touch `modules/tool-delegate/` at all (`git log
3a43b29..HEAD -- modules/tool-delegate/` is empty), so the patch applied
cleanly with `git apply`, zero fuzz, no manual rebase needed.

---

### DIFF
--- a/modules/tool-delegate/amplifier_module_tool_delegate/__init__.py
+++ b/modules/tool-delegate/amplifier_module_tool_delegate/__init__.py
@@ -240,6 +240,104 @@ def _validate_call_budget(value: Any) -> int | None:
     return value or None  # 0 -> None (explicit opt-out)
 
 
+# ---------------------------------------------------------------------------
+# SPAWN-BUDGET treatment (LOCAL PATCH -- not part of upstream
+# amplifier-foundation). See settings.max_spawns_per_session in this
+# module's docstring above. Distinct from the Layer 1 LLM-call budget
+# above: this caps the *count* of new sub-session spawns, not the number
+# of LLM calls within any one of them, and -- unlike
+# ``_validate_call_budget`` -- a bad value never raises. A misconfigured
+# spawn budget must never take down the whole delegate tool; it degrades
+# to unlimited (today's behavior) instead, with a logged warning.
+# ---------------------------------------------------------------------------
+
+
+def _parse_spawn_budget(value: Any) -> int | None:
+    """Defensively parse ``settings.max_spawns_per_session``.
+
+    Returns the parsed non-negative int budget, or ``None`` for
+    "unlimited" (no cap -- today's behavior, byte-identical when this
+    setting is absent). ``None`` is also returned -- with a logged
+    warning, never an exception -- for any value that cannot be
+    confidently parsed as a non-negative integer, so a typo or a bad
+    config value degrades to unlimited rather than either silently
+    blocking every new spawn or crashing tool construction.
+
+    Accepts:
+    - ``None`` -> ``None`` (unlimited; the default)
+    - ``int`` (not ``bool``) -> itself, when >= 0
+    - ``str`` -> ``int(value.strip())``, when that parses to a
+      non-negative int (e.g. ``"3"``, ``" 3 "``)
+
+    Rejects (warns, falls back to ``None`` / unlimited):
+    - ``bool`` (``bool`` is an ``int`` subclass in Python;
+      ``True``/``False`` must never silently become ``1``/``0`` here)
+    - a negative int, or a string that parses to a negative int
+    - ``float``, ``list``, ``dict``, or any other non-str/int type
+    - a string that does not parse as a base-10 integer (e.g. ``"lots"``)
+    """
+    if value is None:
+        return None
+    if isinstance(value, bool):
+        logger.warning(
+            "max_spawns_per_session must be a non-negative integer (or a "
+            "numeric string), not a bool: %r. Falling back to unlimited "
+            "spawns.",
+            value,
+        )
+        return None
+    if isinstance(value, int):
+        if value < 0:
+            logger.warning(
+                "max_spawns_per_session must be >= 0, got %d. Falling back "
+                "to unlimited spawns.",
+                value,
+            )
+            return None
+        return value
+    if isinstance(value, str):
+        try:
+            parsed = int(value.strip())
+        except ValueError:
+            logger.warning(
+                "max_spawns_per_session could not be parsed as an integer: "
+                "%r. Falling back to unlimited spawns.",
+                value,
+            )
+            return None
+        if parsed < 0:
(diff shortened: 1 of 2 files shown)


bkrabach:
All CI checks green (6/6 Tests jobs across ubuntu/windows x Python 3.11/3.12/3.13, plus license/cla). This PR is self-authored by a repo maintainer/admin; merging via admin override at the requesting user's explicit direction since branch protection requires a review this account cannot self-provide. No checks were bypassed -- all required status checks passed before merge.

PR #359, opened 2026-09-06
fix(bundle): instruction @mentions lead the system-prompt context block

bkrabach:
## The defect

A root bundle whose instruction is a single `@mention` had its own `context/system.md`
emitted **LAST** in the system prompt, behind every composed bundle's peripheral
awareness include.

Measured on this host, `anchors-amp-dev`, `amplifier run "hi"`:

| | before | after |
|---|---|---|
| `raw.system` chars | 96,898 | **96,898** (unchanged) |
| `<context_file>` blocks | 23 | **23** (unchanged) |
| first block | `gitea:context/gitea-awareness.md` | **`@anchors-amp-dev:context/system.md`** |
| byte offset of `"You are Amplifier"` | **72,491** (74.8% in) | **225** |
| `mentions:resolved` | 23 resolutions, 0 failed | 23 resolutions, 0 failed |

The mention always resolved correctly. Only its **placement** was wrong.

## Root cause

`_create_system_prompt_factory` added `bundle.context` includes to the
`ContentDeduplicator` **before** resolving the instruction's `@mentions`, and
`format_context_block` emits in deduplicator **insertion order**.

**Gate verified by reading, not assumed:** `format_context_block` iterates
`deduplicator.get_unique_files()`, which iterates `self._content_by_hash.items()` —
a plain dict, insertion-ordered, no sort anywhere on the path
(`amplifier_foundation/mentions/loader.py:42`, `mentions/deduplicator.py:63-70`).
So the swap is sufficient; no explicit ordering fix inside `format_context_block`
was needed.

## The change

One reordering in `_create_system_prompt_factory`: resolve the instruction's
`@mentions` first, then add the `context:` includes. **No content is inlined into
`main_instruction`** — the `<context_file>` block contract is untouched, only block
order moves.

## Semantics that deliberately DID change

`mentions:resolved` keeps identical **set** semantics (same resolutions, same
`failed` list, same `deduplicated_count`) but the **order of the `resolutions`
list** now leads with the instruction's mentions. Flagging it here rather than
letting a consumer discover it.

For a file referenced **both** as a `context:` include and an instruction mention:
still emitted **once**; its `paths=` label now **leads** with the author's explicit
`@mention`, with the include's own attribution trailing (no attribution is dropped).

## Tests

`tests/test_instruction_mentions_lead_context.py` — 4 tests, POSIX + Windows clean
(tmp_path and `pathlib` only; no shell, no `skipif`).

**Fail-before / pass-after, run by stashing only `_prepared.py`:**

```
=== ON MAIN CODE ===
FF.F
FAILED test_instruction_mention_is_the_first_context_block
FAILED test_double_referenced_file_emitted_once_mention_leads
FAILED test_mentions_resolved_payload_semantics_unchanged
3 failed, 1 passed
=== ON BRANCH ===
4 passed in 0.23s
```

Stated honestly: **3 fail-before, 1 no-regression guard.**
`test_no_instruction_mentions_output_is_byte_identical` passes on main **by
construction** — it pins that bundles without instruction mentions are untouched
by the reorder. A guard that failed on main would be pinning the bug, not the
invariant.

`test_mentions_resolved_payload_semantics_unchanged` is instructive: on main it
failed only on its **last** line (the ordering assertion). Every assertion before
it — the resolution set, the `failed` list, `deduplicated_count` — **passed on
main**, which is the direct evidence that the event's set semantics are unchanged.

## Byte-identity comparison (main vs branch)

`docs/lanes/mikb-instruction-mentions-lead/evidence/byte_identity_probe.py`, same
script run against both revisions at fixed paths:

```
### MAIN (e681764, context includes first)
A_no_instruction_mentions:  sha256=d7f616c6a3088e55f2a21372213acd87438025eb51abf849cb94015846635c34 len=247
B_with_instruction_mention: sha256=1b5d114da8f515287d011564fee150c3be2d8792271ab404616d8e8f753458c0 len=379 'You are Amplifier'@offset=333
### BRANCH (instruction mentions resolved first)
A_no_instruction_mentions:  sha256=d7f616c6a3088e55f2a21372213acd87438025eb51abf849cb94015846635c34 len=247
B_with_instruction_mention: sha256=25bcb36828224b93624ecfe9719ee9cd4f32144a69c834b3ee3b68bc2db82fec len=379 'You are Amplifier'@offset=123
```

No-mention case: **identical sha256**. Mention case: **identical length** (379 both)
— content is neither added nor dropped, only reordered.

## Full suite

`uv run pytest tests/ -q` → **1787 passed, 1 skipped, 1 failed**.
The one failure, `tests/test_sources.py::TestFileSourceHandler::test_resolve_existing_file`
(`assert PosixPath('/tmp') == PosixPath('/tmp/tmpXXXX')`), **is pre-existing on
`origin/main` on this host** — verified by re-running it with `_prepared.py` stashed.
It is a file-source-root resolution assertion, untouched by this change.

## Finding, reported rather than absorbed

The task premise said the plain `anchors` bundle "has an inline body and no
mentions". **It does not** — `anchors`'s instruction is also
`@anchors:context/system.md`, so it exhibits the same defect and the same fix:
`"You are Amplifier"` moved **72,467 → 201**, with `raw.system` chars (96,344),
block count (23), and `mentions:resolved` (23/0) all unchanged.

### DIFF
--- a/amplifier_foundation/bundle/_prepared.py
+++ b/amplifier_foundation/bundle/_prepared.py
@@ -412,10 +412,33 @@ class PreparedBundle:
             deduplicator = ContentDeduplicator()
 
             # Build mention_to_path map for context block attribution
-            # This includes BOTH bundle context files AND @mentions from instruction
+            # This includes BOTH @mentions from instruction AND bundle context files
             mention_to_path: dict[str, Path] = {}
 
-            # 1. Bundle context files (from context: section)
+            # ORDER IS LOAD-BEARING. Both the deduplicator and mention_to_path are
+            # insertion-ordered dicts, and format_context_block emits in that order
+            # (get_unique_files iterates _content_by_hash). The instruction's own
+            # @mentions are the bundle author's primary voice, so they are resolved
+            # FIRST and therefore lead the context block; composed bundles' context:
+            # includes (peripheral awareness files) follow. Resolving them the other
+            # way round buried a root bundle's system.md behind every composed
+            # bundle's includes -- 73% into a 99 KB prompt in the reported case.
+
+            # 1. Resolve @mentions from main instruction (re-loads files each call)
+            mention_results = await load_mentions(
+                main_instruction,
+                resolver=resolver,
+                deduplicator=deduplicator,
+            )
+
+            # Add @mention results to mention_to_path for attribution.
+            # Done before the context: includes below so that a file referenced both
+            # ways leads its paths= label with the author's explicit @mention.
+            for mr in mention_results:
+                if mr.resolved_path:
+                    mention_to_path[mr.mention] = mr.resolved_path
+
+            # 2. Bundle context files (from context: section)
             # Add to deduplicator and mention_to_path for unified formatting
             for context_name, context_path in captured_bundle.context.items():
                 if context_path.exists():
@@ -425,19 +448,7 @@ class PreparedBundle:
                     # Add to mention_to_path for attribution (context_name → path)
                     mention_to_path[context_name] = context_path
 
-            # 2. Resolve @mentions from main instruction (re-loads files each call)
-            mention_results = await load_mentions(
-                main_instruction,
-                resolver=resolver,
-                deduplicator=deduplicator,
-            )
-
-            # Add @mention results to mention_to_path for attribution
-            for mr in mention_results:
-                if mr.resolved_path:
-                    mention_to_path[mr.mention] = mr.resolved_path
-
-            # 3. Format ALL context as XML blocks (bundle context + @mentions)
+            # 3. Format ALL context as XML blocks (@mentions + bundle context)
             # format_context_block uses deduplicator for unique content and
             # mention_to_path for attribution (showing name → resolved path)
             all_context = format_context_block(deduplicator, mention_to_path)
--- a/docs/lanes/mikb-instruction-mentions-lead/DONE-NOTE.md
+++ b/docs/lanes/mikb-instruction-mentions-lead/DONE-NOTE.md
@@ -1,0 +1,276 @@
+# DONE-NOTE — `model_performance-mikb`
+
+Lane: `mikb-instruction-mentions-lead`
+Repo: `microsoft/amplifier-foundation`, branch `lane/mikb-instruction-mentions-lead`
+Draft PR: <https://github.com/microsoft/amplifier-foundation/pull/359>
+Base: `origin/main` @ `e681764`
+Date: 2026-09-06
+
+## OUTCOME: **A — RESOLVED.** All deliverables DONE. Nothing NOT-POSSIBLE. Cap did not bind.
+
+**Landing stage.** This lane's final state is the draft PR, marked ready. Procedure 4
+forbids merging; the manager re-runs the fail-befores and merges. The live installed
+CLI on this host still imports the unpatched `amplifier_foundation` — that is the
+landing stage, not an incomplete deliverable.
+
+---
+
+## Deliverables
+
+| # | Deliverable | State |
+|---|---|---|
+| 1 | GATE: verify `format_context_block` emits in INSERTION order **by reading it** | **DONE** |
+| 2 | Fail-before unit tests | **DONE** (3 fail-before + 1 no-regression guard — see the honesty note) |
+| 3 | Do NOT inline mention content into `main_instruction` | **DONE** |
+| 4 | `mentions:resolved` semantics unchanged; order change noted explicitly | **DONE** |
+| 5 | REAL-SESSION CHECK on this host (source override, no cache edits) | **DONE** |
+| 6 | ONE amplifier-tester DTU smoke, foundation branch pinned; DTU destroyed, ledger 0 open | **DONE** |
+| 7 | DRAFT PR, marked ready, with a verification comment | **DONE** |
+| 8 | DONE-NOTE.md at the lane artifact root | **DONE** (this file) |
+
+---
+
+## 1. THE GATE (done first, as instructed) — the approach is valid
+
+`format_context_block` does **not** sort. Read, not assumed:
+
+- `amplifier_foundation/mentions/loader.py:42` — `unique_files = deduplicator.get_unique_files()`,
+  then `for cf in unique_files:` at `:56`, joined `"\n\n"` at `:75`. No `sorted()`, no key.
+- `amplifier_foundation/mentions/deduplicator.py:63-70` — `get_unique_files()` is a list
+  comprehension over `self._content_by_hash.items()`, a plain `dict` (`:24`). Python dicts are
+  insertion-ordered (3.7+), and `add_file` (`:43-47`) inserts on first sight of a content hash.
+
+**Conclusion: emission order == deduplicator insertion order.** The swap is therefore
+sufficient, and no explicit ordering fix inside `format_context_block` was required.
+
+## 2. The change
+
+One reordering in `amplifier_foundation/bundle/_prepared.py::_create_system_prompt_factory`:
+resolve the instruction's `@mentions` into the `ContentDeduplicator` **before** adding
+`captured_bundle.context` includes. A comment marks the ordering as load-bearing so a future
+refactor does not silently undo it.
+
+Nothing else moved. `main_instruction` is untouched; no content is inlined; the
+`f"{main_instruction}\n\n---\n\n{all_context}"` shape is unchanged.
+
+## 3. Tests — stated honestly
+
+`tests/test_instruction_mentions_lead_context.py`, 4 tests. POSIX + Windows clean
+(`tmp_path` + `pathlib` only; no shell, no platform branch, therefore **no `skipif`** —
+the gate four PRs in this batch are stuck on does not apply here).
+
+Fail-before proof, produced by stashing **only** `_prepared.py` (the test file stayed):
+
+```
+=== ON MAIN CODE ===
+FF.F
+FAILED test_instruction_mention_is_the_first_context_block
+FAILED test_double_referenced_file_emitted_once_mention_leads
+FAILED test_mentions_resolved_payload_semantics_unchanged
+3 failed, 1 passed in 0.04s
+
+=== ON BRANCH ===
+4 passed in 0.23s
+```
+(raw: `evidence/fail-before-main.txt`)
+
+**3 fail-before, 1 no-regression guard.** The goal asked for three tests that all fail on
+main, one of them being "no instruction mentions → byte-identical output vs main". That one
+**cannot** fail on main: a test asserting output equals main's output passes on main by
+construction. Making it fail would mean pinning the bug rather than the invariant. So it is
+shipped as a guard and the byte comparison is carried out separately, below. Reported rather
+than fudged.
+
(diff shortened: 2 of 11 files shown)


bkrabach:
## Verification

**Fail-before counts** — produced by stashing *only* `amplifier_foundation/bundle/_prepared.py`, leaving `tests/test_instruction_mentions_lead_context.py` in place:

```
=== ON MAIN CODE (e681764) ===
FF.F
FAILED tests/test_instruction_mentions_lead_context.py::test_instruction_mention_is_the_first_context_block
FAILED tests/test_instruction_mentions_lead_context.py::test_double_referenced_file_emitted_once_mention_leads
FAILED tests/test_instruction_mentions_lead_context.py::test_mentions_resolved_payload_semantics_unchanged
3 failed, 1 passed in 0.04s

=== ON BRANCH ===
4 passed in 0.23s
```

**3 fail-before, 1 no-regression guard** (`test_no_instruction_mentions_output_is_byte_identical`, which passes on main by construction — it pins that mention-free bundles are untouched).

Full suite: **1787 passed, 1 skipped, 1 failed**; the one failure (`test_sources.py::TestFileSourceHandler::test_resolve_existing_file`) reproduces identically on `origin/main` on this host and is unrelated.

**Before/after byte offset of `"You are Amplifier"` in `raw.system`** — real `amplifier run "hi"` sessions on the host, source override via `PYTHONPATH` (no `~/.amplifier/cache` edits; the installed `amplifier_foundation` was verified byte-identical to `origin/main` except this one file, so it is an exact one-file A/B):

| bundle | before | after | `raw.system` chars | `<context_file>` blocks | `mentions:resolved` |
|---|---|---|---|---|---|
| `anchors-amp-dev` | **72,491** | **225** | 96,898 → **96,898** | 23 → **23** | 23/0 failed → 23/0 failed |
| `anchors` | **72,467** | **201** | 96,344 → **96,344** | 23 → **23** | 23/0 failed → 23/0 failed |

First `<context_file>` after `---` went from `gitea:context/gitea-awareness.md` to **`@anchors-amp-dev:context/system.md`** (and from `modes:context/modes-instructions.md` to **`@anchors:context/system.md`**). Notify README absent in every run.

**DTU smoke** (`amplifier-tester`, one environment `mikb-foundation-order`, foundation branch pinned via a Gitea `url_rewrites` snapshot). It produced a real before-arm *inside the same DTU* by pinning the mirror to `e681764`, reinstalling, measuring, and restoring:

| bundle (DTU) | mention block position before → after | `"You are Amplifier"` offset | chars | blocks |
|---|---|---|---|---|
| `anchors` | **8 of 8 (LAST)** → **1 of 8 (FIRST)** | 16,930 → **190** | 40,732 → 40,732 | 8 → 8 |
| `anchors-amp-dev` | **9 of 9 (LAST)** → **1 of 9 (FIRST)** | 19,232 → **214** | 43,564 → 43,564 | 9 → 9 |

Both providers wired and exercised (anthropic `claude-sonnet-5`, openai `gpt-5.6-sol`), both bundles on each, all exit 0, no startup regressions. DTU destroyed; lane ledger **0 open rows**; the created Gitea repo deleted (HTTP 204); the shared mirror and two other sessions' DTUs untouched.

Note the DTU's app-bundle set is 2 rather than the host's 19, so its block counts (8/9) and sizes (~41–44k) are **not** comparable to the host's 23 / ~97k. The DTU carries the ordering claim; the host runs carry the magnitudes.

**Finding, reported rather than absorbed:** the task premise said the plain `anchors` bundle "has an inline body and no mentions". It does not — `anchors`'s instruction is also `@anchors:context/system.md`, so it had the same defect and gets the same fix. The mention-free no-regression case is covered by the unit guard and the sha256 comparison in the PR body instead.

Full lane note: `docs/lanes/mikb-instruction-mentions-lead/DONE-NOTE.md`. Evidence: `docs/lanes/mikb-instruction-mentions-lead/evidence/`.

Spend: ~$1.82 against a $3.00 authority (4 host runs @ $0.27 = $1.08; DTU smoke ~$0.74). Cap did not bind; no deliverable was dropped.

bkrabach:
## Manager verification — all six gates re-run by me in a scratch clone. Merging.

Head `1c184b3`, base `e681764` = current `origin/main`. **CI green 7/7** (ubuntu + Windows × py3.11/3.12/3.13). Exactly one source file touched: `amplifier_foundation/bundle/_prepared.py`.

**1. Ordering — read, not assumed.** This was the gate that could have voided the whole approach, so I checked it myself. `deduplicator.py:63-70` returns a comprehension over `self._content_by_hash.items()` — a plain insertion-ordered dict — and there is **no `sort`/`sorted` anywhere** in `loader.py` or `deduplicator.py`. Emission order **is** insertion order, so swapping the two population steps is sufficient and no explicit ordering fix inside `format_context_block` was needed.

**2. Fail-before, run by me:**

```
main   e681764 : 3 failed, 1 passed
branch 1c184b3 : 4 passed
```

Failing on main: `test_instruction_mention_is_the_first_context_block`, `test_double_referenced_file_emitted_once_mention_leads`, `test_mentions_resolved_payload_semantics_unchanged`.

**The "1 passed" is the honest part.** The lane reports plainly that test 2 ships as a **no-regression guard, not a fail-before**, because *a test asserting output equals main's output cannot fail on main* — making it fail would pin the bug rather than the invariant. It carried the byte comparison separately instead, and the evidence file shows it: for the no-instruction-mention case, `sha256=d7f616c6a3088e55…` is **identical on both revisions**, while the with-mention case differs (`1b5d114d…` → `25bcb368…`). That is the right call, reported rather than fudged.

**3. Real session — the number that matters.** `~/.amplifier/cache` is untouched (only an untracked `.amplifier_cache_meta.json`; no tracked file modified), and the override was `PYTHONPATH` against an installed package verified byte-identical to `origin/main` except the one changed file.

| bundle | "You are Amplifier" offset | raw.system | context_files |
|---|---|---|---|
| anchors-amp-dev | **72,491 → 225** | 96,898 → 96,898 | 23 → 23 |
| anchors | **72,467 → 201** | 96,344 → 96,344 | 23 → 23 |

First block moves from `gitea:context/gitea-awareness.md` to `@anchors-amp-dev:context/system.md`; notify README absent. **Total length unchanged** — only order moved, which is exactly the contract.

**4. DTU smoke** ran with the branch pinned; **0 open ledger rows, 0 containers left**. Spend **$1.82 of $3.00**, arithmetic restated at the observed price ($0.74 vs the $1.00 estimate).

**5. `mentions:resolved`** — same resolution set, same `failed` list, same `deduplicated_count` (23/0 both bundles, before and after). Order of `resolutions` changes and is called out explicitly.

**6.** Suite `1787 passed, 1 skipped, 1 failed`; the failure is `test_sources.py::TestFileSourceHandler::test_resolve_existing_file`, which the lane reproduced **identically on `origin/main` with the change stashed** — pre-existing, not absorbed.

### Two findings worth keeping

- **The task premise was FALSE and the lane caught it**: the plain `anchors` bundle does *not* have an inline body — its instruction is also `@anchors:context/system.md`, so it had the same defect and gets the same fix. It only found this by running the before-arm it wasn't asked for ($0.27).
- **`amplifier update` alone does not downgrade the foundation package inside a DTU** — it refreshes the bundle cache while `site-packages` stays at the newer sha, a split state that would silently produce a **false before-arm**. Caught via `direct_url.json`; `uv tool install --reinstall --force` is required. That one is a trap for any future lane pinning a foundation ref.

Squash + `--admin` (CI green; `mergeStateStatus` BLOCKED is the known required-check livelock).

bkrabach:
## Manager verification — all gates re-run by me in a scratch clone. Admin-merging.

Head `1c184b3`, base `e681764` = current `origin/main`. **CI 7/7 SUCCESS** (ubuntu + Windows × 3.11/3.12/3.13). Outside `tests/` and `docs/`, the diff touches exactly one file: `amplifier_foundation/bundle/_prepared.py`.

### Gate 1 first, because it decided whether the approach was valid at all

The goal ordered this first: if `format_context_block` **sorted**, the step-swap would be void. I read it rather than trusting the marker:

- `mentions/loader.py:42` — `unique_files = deduplicator.get_unique_files()`, iterated with **no `sorted()`**
- `mentions/deduplicator.py:69` — comprehends over `self._content_by_hash.items()`, a **plain insertion-ordered dict**

**Emission order == insertion order**, so swapping the two steps is sufficient and no explicit ordering fix inside `format_context_block` was needed. The lane's reading is correct.

### Gate 2 — fail-before, run by me on both trees

```
main   e681764 : 3 failed, 1 passed
branch 1c184b3 : 4 passed
```

Failing by name on main: `test_instruction_mention_is_the_first_context_block`, `test_double_referenced_file_emitted_once_mention_leads`, `test_mentions_resolved_payload_semantics_unchanged`.

**The "1 passed" is an honest deviation, not a gap.** A test asserting *"output is byte-identical to main"* **cannot fail on main by construction** — making it fail would pin the bug rather than the invariant. It ships as a no-regression guard, with the byte comparison carried separately (identical `sha256 d7f616c6…` on both revisions for the no-mention case). Reported, not fudged.

### Gate 3 — the real session, which is where this defect actually lives

| | before | after |
|---|---|---|
| `offset_of_You_are_Amplifier` | **72,491** | **225** |
| first `<context_file>` | `gitea:context/gitea-awareness.md` | `@anchors-amp-dev:context/system.md` |
| `raw.system` chars | 96,898 | **96,898** |
| `context_file` blocks | 23 | **23** |
| notify README present | false | false |

**The prompt is the same size and has the same blocks — only the order moved.** That is exactly the intended change, and the unchanged char count is the strongest evidence nothing was dropped or duplicated.

**A false premise in my own item, found and corrected by the lane:** I wrote that the plain `anchors` bundle *"has an inline body and no mentions"*. It does not — its instruction is also `@anchors:context/system.md`, so **`anchors` had the same defect and gets the same fix: 72,467 → 201.** The lane ran that bundle twice (before *and* after) rather than the specified once, which is what turned "no regression" into a measured claim and exposed the bad premise. Worth the extra $0.27.

### Gate 5 — `mentions:resolved` semantics

`resolutions=23 failed=0 dedup=0` on **both** sides, for **both** bundles. Only `first_resolution_mention` changes (`gitea:…` → `@anchors-amp-dev:…`), and the order shift is called out explicitly in the PR body rather than left for a reader to discover.

### Gates 4 and 6

DTU smoke **$0.74** observed against a $1.00 estimate; total spend **$1.82 of $3.00**, arithmetic closes. Suite `1787 passed, 1 skipped, 1 failed` — the one failure (`tests/test_sources.py::TestFileSourceHandler::test_resolve_existing_file`) **reproduces identically on `origin/main` with the change stashed**, so it is pre-existing and was not absorbed. No `skipif` needed: the new tests use `tmp_path` and `pathlib` only, no shell, no platform branch — and Windows CI confirms it.

### One finding worth keeping for future lanes

> `amplifier update` alone does **not** downgrade the foundation Python package inside a DTU — it refreshes the bundle cache while `site-packages` stays at the newer sha, a split state that would silently produce a **false before-arm**. `uv tool install --reinstall --force` is required.

Caught via `direct_url.json`. That is a trap any future lane pinning a foundation ref in a DTU would have fallen into.

Squash + `--admin`, per the owner's standing directive for this defect family.

PR #362, opened 2026-09-06
fix(foundation): four hygiene defects -- session-analyst glob, two commit footers, dead routing, stale <example> claim

bkrabach:
## Summary

Four upstream-hygiene defects in foundation's own agent/context files, found by
`_dashboard/foundation-pull-review.md` **while deciding what NOT to pull into the anchors family**.
`ux32` has since landed (`200dfe6`), so **none of this is load-bearing for anchors** — the consumer
is foundation's own sessions. Fixed small, with a fail-before test on each.

| # | Defect | Fix |
|---|---|---|
| 1 | `session-analyst` script glob never matched at runtime | glob + the test that pinned the broken form |
| 2 | Two commit footers in one session; a `sources:` example that overrides nothing | one footer repo-wide; `modules:` level + round-trip test |
| 3 | `ecosystem-expert` routed to three agents in no roster, with no delegate tool | section deleted |
| 4 | `bundle-design-expert` required `<example>` blocks its own cited source bans | both lines corrected; awareness → one-line pointer |

---

## (1) The glob missed the real cache directory

```bash
find ~/.amplifier/cache -path '*/amplifier-foundation/scripts/amplifier-session.py'    # EMPTY
find ~/.amplifier/cache -path '*/amplifier-foundation-*/scripts/amplifier-session.py'  # HIT
```

The script lives under the hashed module cache. Fixed to `*/amplifier-foundation*/scripts/…`,
which matches **both** the cache layout and a plain source checkout (3 occurrences).

**`tests/test_session_analyst_md.py` pinned the broken string** in `SCRIPT_DISCOVERY_LINE` and in
`test_script_discovery_uses_find`. A test that pins a broken form is worse than no test: it makes
the defect look guarded. Both updated, and `TestScriptDiscoveryResolvesRealLayouts` now resolves
*every* discovery glob in the file against both real layouts via `fnmatch` (`find -path` semantics),
so a glob that compiles but matches nothing fails loudly.

### Real-session check — a real delegation, both directions

Scratch `AMPLIFIER_HOME`, scratch venv. **`~/.amplifier/cache` was never edited** (verified after —
it still carries the broken form); the diff was applied to the *scratch* cache.

| | released agent | patched agent |
|---|---|---|
| resolved | nothing — timed out at 30s, scoped retries empty | `…/cache/skills/amplifier-foundation-c909465861f9d6ce/scripts/amplifier-session.py` |
| script ran | **no** — "neither info nor --help ever executed" | **yes** — `usage: amplifier-session [-h] {diagnose,repair,rewind,info,find}` |

### The `:42` IDENTITY NOTICE: deleted, measured not assumed

The item said to check whether foundation's own composition also excludes `tool-delegate`. It does:

- `behaviors/agents.yaml:31` — `settings.exclude_tools: [tool-delegate]`
- root `bundle.md` includes `foundation:behaviors/agents`
- module default is the same anyway: `tool-delegate/__init__.py:636`, `settings.get("exclude_tools", ["tool-delegate"])`
- `session-analyst`'s own `tools:` declares filesystem/search/bash — it does **not** re-declare
  delegate, which is the only carve-out that would restore it (`session_spawner.py::_filter_tools`)

So a "don't delegate to yourself" warning is unreachable text. `TestNoSelfDelegationNotice` pins
both invariants, so the notice returns if either changes.

Also: the `events.jsonl` rule went from 5 statements to 1 (`404 → 374` lines).

---

## (2) One commit footer, and an example that overrode nothing

`common-agent-base.md` is @mentioned by every foundation agent and, via `common-system-base.md`, by
the root system prompt — everything wrong in it is paid on every spawn.

- **Two footers in one session.** This file mandated the emoji form while bundle system prompts
  mandate the plain `Generated with Amplifier`. Owner decision: the plain form. This file is now the
  single place it is stated; `agents/git-ops.md` (×2) and `context/ISSUE_HANDLING.md` (×1) reference
  it instead of restating it.
- **The emoji contradiction** at `:47` — "Use emojis only when the user explicitly requests them",
  15 lines under a mandate to emit one unprompted. Resolved by the above.
- **The todo absolute** at `:200` deleted — it restated a system-prompt operating rule as an
  absolute, in a file loaded by agents that may have no todo tool.
- **The `sources:` example** omitted the `modules:` level. The real loader reads
  `settings.get("sources", {}).get("modules", {})`, so a flat mapping parses fine and overrides
  **nothing**.

Measured against the real loader in a scratch venv (`app-cli @ c120a365`):

```
fail-before: Loader returned {}, doc promised {'tool-bash': ..., 'provider-anthropic': ...}
pass-after:  12 passed
```

`test_flat_example_would_have_failed` pins the flat shape returning `{}`, so the guard cannot rot
into a tautology.

**On the skip:** the round-trip skips where `amplifier-app-cli` is absent, **with the reason stated
in the skip message**. It cannot become a dev-dependency — app-cli's metadata declares
`Requires-Dist: amplifier-foundation`, so it would be circular. CI's coverage of the same defect is
`TestSourcesOverrideExampleShape`, which always runs and pins the loader's read path as a verbatim
quote.

---

## (3) Dead routing deleted — and a cost flagged, not touched

`behaviors/amplifier-dev.yaml` is the only roster composing `foundation:ecosystem-expert`, and its
`agents.include` is exactly `[foundation:ecosystem-expert]`. `amplifier:amplifier-expert`,
`core:core-expert` and `foundation:foundation-expert` resolve nowhere in it — and the agent has no
delegate tool regardless.

**FLAGGED FOR THE MAINTAINER, NOT CHANGED — `ISSUE_HANDLING` cost.** Re-measured on this tree
(@mention closure, ~4 chars/token):

| agent | files | tokens/spawn |
|---|---|---|
| `ecosystem-expert` | 11 | **~31,384** — `ISSUE_HANDLING.md` alone ~8,576 |
| `amplifier-dev-expert` (post-ux32) | 4 | **~4,810** |

≈ **6.5×**. The item quoted 8,561 for `amplifier-dev-expert`; ux32 has since made it a thin layer,
so the gap is *wider* than reported. Restructuring is explicitly the maintainer's call — untouched.

---

## (4) An agent that contradicted the file it cites

`:232` listed `<example>` blocks as the HOW requirement and `:243` listed "Missing `<example>`
blocks" as a defect — while `:100` @mentions `description-authoring-principles.md`, whose Example
policy reads **"No `<example>` blocks. Not 'at most 2' — zero."** Both lines now point at the
decision-rule-in-WHEN-clause form (V6).

`context/bundle-design-awareness.md`: 51 lines → 3 (~540 → ~63 tokens) off every root prompt
composing `behaviors/bundle-design.yaml`. Nothing lost — the lifecycle, all four recipes and their
required context keys are already in `bundle-design-expert.md:358-369`, which is the context sink
`behaviors/bundle-design.yaml`'s own comment says it should be.

---

## Test plan

Each premise is re-derived from its source rather than hardcoded, so every guard fails loudly if the
composition or the principles file changes.

- [x] `tests/test_session_analyst_md.py` — 2 failed → 43 passed
- [x] `tests/test_common_agent_base_md.py` — 8 failed → 12 passed (real loader, scratch venv)
- [x] `tests/test_agent_docs_hygiene.py` — 5 failed → 9 passed
- [x] Real-session delegation, fail-before **and** pass-after, hashed cache dir confirmed
- [x] Full suite, honest baseline comparison:

| tree | result |
|---|---|
| `origin/main` (`5bc2ed1`, clean worktree) | **1 failed**, 1798 passed, 1 skipped |
| this branch | **1 failed**, 1820 passed, 3 skipped |

Same single failure both sides — `tests/test_sources.py::TestFileSourceHandler::test_resolve_existing_file`,
pre-existing and **not absorbed**. +22 tests, zero new failures. The 2 new skips are the round-trip,
skipped with a stated reason.

**Honest note:** the second pre-existing failure I was told to expect
(`test_grpc_adapter_main.py::TestVerifyModuleType::test_non_isinstance_object_with_mount_passes`) did
**not** reproduce on `origin/main` on this host. Not claiming it as pre-existing; not claiming I
fixed it.

**Windows CI:** no new platform-dependent tests. The only new skip is dependency-based, with its
reason in the message.

## Not done on purpose

- `bundles/anchors/`, `bundles/anchors-amp-dev/` — untouched (`ux32` owns them).
- `ecosystem-expert`'s `ISSUE_HANDLING` structure — flagged above, maintainer's call.
- `experiments/` — excluded from the repo-wide footer sweep as a sandbox outside the live
  composition; one emoji footer survives at
  `experiments/behavioral-anchor-amplifier-dev/context/amplifier-dev/dev-workflows.md:155`.
- `behaviors/amplifier-dev.yaml`'s YAML comment still says "Delegate to foundation-expert" — same
  dead-routing class as (3), but it is a comment (zero token cost) and out of the item's scope.

Full evidence, decisions and spend: `docs/lanes/ezze-foundation-hygiene/DONE-NOTE.md`.

Generated with Amplifier


bkrabach:
## Manager verification — all 6 gates re-run by me in a scratch clone. Merging.

Head `595c245`, base `200dfe6`. **CI 7/7 SUCCESS** (ubuntu + Windows × 3.11/3.12/3.13).

| gate | decisive fact |
|---|---|
| **1. glob** | against the real cache: **OLD form → 0 hits**, **NEW form → 2 hits**. Fail-before on the pin: **5 failed, 38 passed** at main → **43 passed** on branch |
| **2. `:42` notice** | **MEASURED, not argued**: `behaviors/agents.yaml:32` sets `exclude_tools: [tool-delegate]`; notice now **0** occurrences |
| **3. footer** | emoji form **0 files**, plain `Generated with Amplifier` **4 files**; `:200` todo absolute **0**; `sources:` example now carries the `modules:` level (`:150`) — round-trip suite **20 passed, 3 skipped** |
| **4. dead routing** | `amplifier:amplifier-expert\|core:core-expert\|foundation:foundation-expert` in ecosystem-expert.md → **0** |
| **5. bundle-design** | `Missing <example>` → **0**; `bundle-design-awareness.md` is now **3 lines** |
| **6. suite** | **1941 passed, 2 failed** — and I ran those two at `origin/main`: **both fail there identically**. Waived on measurement, both sides quoted |

### The judgment call I most wanted checked, and it was made correctly

Deliverable (1) told the lane to delete the IDENTITY NOTICE *only if* foundation's own composition excludes `tool-delegate` — otherwise keep it. It **measured** rather than assumed, and traced the whole chain: `behaviors/agents.yaml:31` sets `exclude_tools: [tool-delegate]`, the root bundle includes that behavior, `tool-delegate/__init__.py:636` defaults the same, and session-analyst does not re-declare delegate — the only carve-out per `session_spawner.py::_filter_tools`. That is a deletion resting on four measured facts, not on a plausible reading.

### One honest correction to the lane's own report

Its DONE-NOTE says `test_grpc_adapter_main.py::…test_non_isinstance_object_with_mount_passes` **did not** reproduce on `origin/main` on its host, and it carefully declined to claim it as pre-existing *or* as fixed. **On my host it does reproduce at `origin/main`** — I ran exactly those two node-ids on a clean main worktree and got `2 failed`. So it is pre-existing after all. Recording this because the lane's refusal to over-claim is what made the discrepancy visible enough to settle.

### Deviations, all sound

- Fixed **three** footer restatements the item did not name (`agents/git-ops.md` ×2, `context/ISSUE_HANDLING.md` ×1) — required, since (2) says *exactly one form repo-wide*. ISSUE_HANDLING's structure and token cost untouched, per the scope-out.
- **`experiments/` deliberately excluded** from the footer sweep, with the one surviving emoji footer **named** (`experiments/behavioral-anchor-amplifier-dev/context/amplifier-dev/dev-workflows.md:155`) rather than silently swept. That directory is now `6phe`'s subject.
- **amplifier-app-cli not added as a dev dependency** — its metadata declares `Requires-Dist: amplifier-foundation`, so it would be circular. The round-trip test skips with that reason **in the skip message**, and CI still covers the defect via the always-running `TestSourcesOverrideExampleShape`.
- ISSUE_HANDLING cost **flagged, not restructured**, as instructed — re-measured at ~31,384 tok/spawn vs amplifier-dev-expert's ~4,810 post-ux32 (**~6.5×**, wider than the item's quoted figure).

### Goal defect it reported against me, and it is right

My goal stated the cap as `0 runs × 0 arms × $0` **while also mandating a real `amplifier run`** — the arithmetic does not close on its own terms. It is a wording defect, not a funding gap (the carve-out made the work reachable), and the correct statement would read `2 runs × 1 arm × ~$0.85 / 1.00 = $1.70`. Actual: **$1.72**. It also ran **two** sessions rather than one, because *"a pass-after with no fail-before proves nothing about a glob that compiles either way"* — correct, and worth the extra dollar.

PR #364, opened 2026-09-06
fix(recipes): validate-agents discovers only 23 of 32 agents -- widen discovery to repo-wide scope

bkrabach:
Closes the discovery gap `6phe` named as its finding F3.

## The problem

`recipes/validate-agents.yaml`'s `agent-discovery` step walked a hand-maintained list of three directories. `recipes/validate-bundle-repo.yaml` and `tests/test_anchors_bundles_dry.py`'s `_agent_files()` both glob the repo. On this repo: **23 files vs 32.**

That asymmetry is the root cause of a three-sweep recurrence chain — 11 `<example>` violators under `experiments/**/agents/` survived **#341** (`1b16a7d`, swept root `agents/`) and **ux32** (`200dfe6`, swept `bundles/**`) because the recipe most people run never looked there. And it reported a clean `"23 found, 3 locations"` the whole time, with no signal that other agent directories existed at all.

## The change

Discovery now globs `rglob("*/agents/*.md") + glob("agents/*.md")` and drops excluded parts, using **`6phe`'s existing `AGENT_SCAN_EXCLUDED_PARTS` verbatim** (`tests/test_anchors_bundles_dry.py:52-69`) — not a second list.

- `discovery_results` gained `location_counts`, `scan_patterns`, `excluded_parts`
- `location` is **derived** from each file's parent directory, so a new `agents/` tree names itself the first time it appears
- `synthesize-report` and `quick-approval` must now print a coverage block (patterns, exclusions, total, per-location counts)
- **Discovery only.** No check, threshold, severity or verdict rule changed. On a repo whose agents all live in the three old locations, output is identical.

## Before / after — both measured by running the recipe

| | BEFORE | AFTER |
|---|---|---|
| version | v1.5.1 | v1.6.0 |
| `run_id` | `run-c26b396f2f1e` | `run-67388a546d75` |
| discovered | **23** | **32** |
| locations | 3 | 6 |
| **verdict** | **PASS** | **FAIL** |

Newly walked: `context/agents/` (4), `examples/agents/` (1), `experiments/build-up/agents/` (4). The 32 files match `6phe`'s `evidence/scope-parity.txt` set exactly — all three scopes now agree.

## ⚠️ The verdict did NOT stay PASS — and it is reported, not fixed

```
4x ERROR   NO_FRONTMATTER   context/agents/{delegation-instructions,multi-agent-patterns,
                                            session-repair-knowledge,session-storage-knowledge}.md
1x WARNING NO_TOOLS_SECTION examples/agents/file-responder.md
0 findings  experiments/build-up/agents/  (4 files -- the tree that started the chain is clean)
```

**These are not dirty agents.** The four are *context documents* in a directory named `agents/` — loaded as `context:` by `behaviors/agents.yaml:44-45` and `behaviors/tasks.yaml:18-19`, `@`-mentioned by `agents/session-analyst.md:369-370`. Nothing spawns them, so `NO_FRONTMATTER`'s "agent won't load" describes an event that cannot occur. The defect is the glob classifying by *directory name*.

Per the item's scope-out: **reported, left unfixed, and the recipe was not weakened to restore PASS.** The correct fix (discover by path, classify by `meta:` presence — matching the loader's own contract) is a separate item.

## Shared-helper question — **declined**, with a cheap substitute built instead

A shared discovery implementation is not cheap: recipe steps are self-contained heredocs with no import mechanism, and routing discovery through `amplifier_foundation` would make *scope* depend on an import the recipe explicitly allows to be missing (its own `structural-validation` step degrades to a no-op on `ImportError`).

Built instead: **`TestDiscoveryScopeParity`** — compares the exclusion-set literals across both recipes and the guard, and **executes the recipe's own discovery step** to prove the file sets match (a test that re-implemented the glob would pass while the recipe diverged).

**Fail-before:** both guards fail against the unmodified recipe, naming all 9 missing files. **Pass-after:** `8 passed` (was 6).

## Verification

- Suite: `2 failed, 2017 passed, 3 skipped` — the two are the known pre-existing failures (`test_sources`, `test_grpc_adapter_main`), untouched here
- `validate-bundle-repo` deliberately **not** run, so `6phe`'s F1 artifact-rewrite side effect could not fire; `git status` confirms only the two source files changed

Full record + evidence: `docs/lanes/xe1u-validate-agents-discovery/`

Draft — **do not merge**; the manager re-runs the before/after counts.

### DIFF
--- a/docs/lanes/xe1u-validate-agents-discovery/DONE-NOTE.md
+++ b/docs/lanes/xe1u-validate-agents-discovery/DONE-NOTE.md
@@ -1,0 +1,322 @@
+# Lane xe1u -- `validate-agents` discovery widened (23 -> 32)
+
+**Item:** `model_performance-xe1u` (project `model_performance`)
+**Branch:** `lane/xe1u-validate-agents-discovery` off `origin/main` @ `5a9e07b`
+**Outcome:** **A. RESOLVED** -- every deliverable DONE. Nothing was cut by the cap.
+**Landing stage:** draft PR only. Per the goal's LANDING STAGE clause, a deliverable
+whose final state needs a merge is satisfied here as *demonstrated fail-before /
+pass-after and shipped for landing*; the merge is the manager's next stage.
+
+**Headline:** widening discovery flipped the verdict **PASS -> FAIL**, and the goal
+anticipated exactly this. The four newly-surfaced ERRORs are **not dirty agents** --
+they are four *context documents* that live in a directory named `agents/`. Reported,
+**left unfixed**, per the scope-out. The recipe was **not** weakened to restore PASS.
+
+---
+
+## Deliverables
+
+| # | Deliverable | State |
+|---|---|---|
+| 1 | Discovery widened to every `**/agents/*.md`, using 6phe's existing exclusion set (quoted) | **DONE** |
+| 2 | Before/after discovered counts, both measured by running the recipe | **DONE** -- 23 -> **32** |
+| 3 | Verdict stays PASS | **DONE as reported: it did NOT.** PASS -> FAIL, headline finding F1, left unfixed |
+| 4 | Report metadata names search locations AND discovered total | **DONE** |
+| 5 | Shared-helper question answered either way | **DONE** -- shared implementation **DECLINED with reason**; a cheap *parity guard* implemented instead |
+| 6 | Suite green, draft PR, not merged | **DONE** |
+| 7 | DONE-NOTE at the lane artifact root | **DONE** (this file) |
+
+---
+
+## 1. Discovery widened -- and where the exclusion set was read from
+
+`recipes/validate-agents.yaml`, step `agent-discovery`, previously walked a
+hand-maintained list of three places (`agents/`, `behaviors/*/agents/`,
+`bundles/*/agents/`). It now globs the repo and drops excluded parts:
+
+```python
+candidates = list(path.rglob("*/agents/*.md")) + list((path / "agents").glob("*.md"))
+```
+
+**The exclusion set was copied, not invented.** Quoting the source verbatim --
+`tests/test_anchors_bundles_dry.py:52-69` (added by lane 6phe):
+
+```python
+# The validator's own agent-discovery scope, copied deliberately rather than
+# approximated. `@foundation:recipes/validate-bundle-repo.yaml`'s
+# `agent-description-validation` step globs
+# `rglob("*/agents/*.md") + glob("agents/*.md")` and drops any path containing
+# one of these parts. Two validators disagreeing about which files exist is
+# exactly how the 11 `experiments/` violators survived #341 and ux32:
+# `validate-agents` discovers 23 agents and never walks `experiments/`, while
+# `validate-bundle-repo` discovers 45 and does. This guard follows the wider
+# one. `docs` is added on top of the validator's set: `docs/` holds frozen lane
+# records that quote violations verbatim as evidence and must never be rewritten.
+AGENT_SCAN_EXCLUDED_PARTS = {
+    "test-fixtures",
+    "tests",
+    "node_modules",
+    ".git",
+    ".venv",
+    "docs",
+}
+```
+
+That six-element set is now `EXCLUDED_PARTS` in the recipe, byte-for-byte, with the
+provenance in a comment above it. **No second list was created** -- and
+`TestDiscoveryScopeParity` (below) fails the build if one ever appears.
+
+## 2. Before / after -- both measured by running the recipe
+
+| | BEFORE | AFTER |
+|---|---|---|
+| Recipe version | v1.5.1 (unmodified) | v1.6.0 (this branch) |
+| `run_id` | `run-c26b396f2f1e` | `run-67388a546d75` |
+| `total_count` | **23** | **32** |
+| `search_locations` | 3 | 6 |
+| `quality_level` | `good` | `critical` |
+| **Verdict** | **PASS** | **FAIL** |
+
+Per-location counts, from the AFTER run's `location_counts`:
+
+| Location | Agents | |
+|---|---|---|
+| `agents/` | 16 | walked before |
+| `bundles/anchors/agents/` | 6 | walked before |
+| `bundles/anchors-amp-dev/agents/` | 1 | walked before |
+| `context/agents/` | 4 | **newly walked** |
+| `examples/agents/` | 1 | **newly walked** |
+| `experiments/build-up/agents/` | 4 | **newly walked** |
+
+**32 is measured, not assumed** -- and it is the *same 32 files*, by path, as
+6phe's `evidence/scope-parity.txt` ("validator scope ... 32 agents / guard (c)
+scope ... 32 agents / identical sets: True"). All three scopes now agree.
+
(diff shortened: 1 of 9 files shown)


bkrabach:
## Manager verification — **NOT merging.** The FAIL is an ARTIFACT, and Windows CI is red on the new parity test.

Head `038801f`, base `5a9e07b`. Gates 1-3 hold; **4 and 5 do not**.

| gate | result |
|---|---|
| **1. counts/verdicts** | 23/3 locations (PASS, `run-c26b396f2f1e`) → **32/6 locations (FAIL, `run-67388a546d75`)**. Newly walked: `context/agents/` 4, `examples/agents/` 1, `experiments/build-up/agents/` 4 |
| **2. exclusion set** | ✅ no divergent list — the recipe cites `AGENT_SCAN_EXCLUDED_PARTS` (`tests/test_anchors_bundles_dry.py:52-69`) and `TestDiscoveryScopeParity` guards the agreement |
| **3. not weakened** | ✅ zero threshold/severity/skip changes in the diff. The only such line is a comment: *"Discovery only. No check, threshold, severity or verdict rule"* |
| **4. new findings** | ❌ **0 real defects, 4 artifacts + 1 benign** |
| **5. CI** | ❌ **Windows red on all three Python versions** |

### Gate 4 — I read the four files myself rather than take the classification on trust

All four ERRORs are `NO_FRONTMATTER` in `context/agents/*.md`. Measured:

```
delegation-instructions.md    frontmatter=0
multi-agent-patterns.md       frontmatter=0
session-repair-knowledge.md   frontmatter=0
session-storage-knowledge.md  frontmatter=0

behaviors/agents.yaml:44-45   - foundation:context/agents/delegation-instructions.md
behaviors/tasks.yaml:18-19    - foundation:context/agents/multi-agent-patterns.md
```

They are **context documents loaded via `context:`** — nothing ever spawns them. They have no frontmatter because they are **not agents**. The discovery predicate classifies by **directory name**, so `context/agents/` collides on the word "agents".

**`experiments/build-up/agents/` — the 4 files this whole chain was about — came back CLEAN.** The one WARNING (`examples/agents/file-responder.md`, no `tools:`) is a fixture demonstrating capability inheritance.

So widening discovery surfaced **zero real defects**. The verdict flip is entirely the collision. **Merging it would install a permanently-red baseline that is wrong** — and a validator that cries wolf gets muted, which costs more than the gap it closed.

### Gate 5 — and Windows shows the widened discovery finds *nothing*

```
tests/test_anchors_bundles_dry.py::TestDiscoveryScopeParity::test_validate_agents_discovers_exactly_the_guards_agent_files
E  AssertionError: In recipe not guard: []; in guard not recipe: {'agents/bug-...pert.md', ...}
E  assert set() == {...}
```

The recipe's discovery returns **`set()`** on Windows while the guard finds the real files — a path-separator mismatch in the glob. **This is not merely a test failure: on a Windows checkout the widened recipe would discover 0 agents and PASS vacuously** — strictly worse than the 23-agent under-count it replaces. Ubuntu 3.11/3.12/3.13 are green; the parity guard did its job by catching this, which is the strongest argument for keeping it.

### What this lane got right, and it is most of it

The measurement is exactly what was asked for, and **it did not weaken the recipe to keep PASS** — it reported the flip and left the cause unfixed, which is why the artifact is visible at all.

**Its F2 is the finding I most want kept**: `validate-bundle-repo`'s reassuring `agents_checked: 32, errors: []` is **quietly wrong on the same files** — `extract_description()` returns an empty string when frontmatter is absent, so a file with **no** frontmatter passes clean. Its count was inflated by 4 (true agent count **28**), and **it cannot detect a broken frontmatter at all.** I quoted that very line as evidence when verifying `6phe` last cycle; it was a validator that structurally could not see this class of problem. Filed separately.

Also honest: **F4 — a FAILing repo costs MORE LLM steps than a PASSing one** (3 vs 2, the conditional deep-analysis phases fire), which reverses once the collision is fixed.

### To merge, in order

1. **Classify by `meta:` presence, not directory name** — match the loader's own contract, so `context/agents/*.md` is correctly not-an-agent. That alone should restore PASS honestly at ~28 agents.
2. **Fix the glob for Windows** so parity holds on all platforms.

Filed as `model_performance-jd97` (discovery scope-awareness, incl. the Windows glob) and `model_performance-vg2b` (the validate-bundle-repo blind spot). Left draft, nothing reverted — the work stands, it needs the classifier before the coverage is worth having.

bkrabach:
**Correction to my previous comment — the follow-up item ids.**

I wrote `model_performance-jd97` and `model_performance-vg2b` before filing them; the tracker assigned different ids. The real ones:

- **`model_performance-39z0`** — classify agents by `meta:` presence rather than directory name (so `context/agents/*.md` is correctly not-an-agent), **and** fix the glob so discovery is not empty on Windows. Acceptance requires the verdict to return to PASS *without* any threshold, severity, skip or exclusion-set change, and forbids deleting the parity guard that caught the Windows defect.
- **`model_performance-dfni`** — `validate-bundle-repo`'s `extract_description()` returning `""` on missing frontmatter, so a frontmatter-less file passes clean and `agents_checked` is inflated.

Both name this branch (`038801f`) as the base to build on rather than starting over, and both are sequenced so `39z0`'s classifier lands first — it determines which files are agents at all, which `dfni`'s count depends on.

Nothing else in the previous comment changes.

bkrabach:
Closing as **superseded by #365**, which merged as `e495755`.

Every commit on this branch is in main — #365 was branched from this head (`038801f`) rather than starting over, so the widened discovery, the report metadata (`location_counts`/`scan_patterns`/`excluded_parts`) and `TestDiscoveryScopeParity` all landed. Content-verified on main after the squash: **20** matches for this PR's discovery/metadata markers, **16** for #365's classifier and env-path additions.

To be explicit about what this PR got right, since it is closed rather than merged:

- The **measurement was correct and is the reason the chain resolved**: widening 23 → 32 is what exposed both the classifier collision and the Windows defect. Neither was visible before.
- It **did not weaken the recipe to keep PASS** — it reported the flip and left the cause unfixed, which is the only reason the artifact was diagnosable rather than baked in as a red baseline.
- Its own ** caught the Windows failure** that would otherwise have shipped as a vacuous PASS discovering 0 agents.
- Its **F2** — `validate-bundle-repo`'s `extract_description()` returning `""` on missing frontmatter, so a frontmatter-less file passes clean — is now `model_performance-dfni`, and it corrected a figure I had quoted as evidence in an earlier verification.

The only thing that could not merge was the verdict state: 4 artifact ERRORs from directory-name classification, plus discovery returning `set()` on Windows. #365 fixed both — the Windows cause turned out to be a repo path interpolated into Python **source**, where `\a` becomes BEL, not a path-separator problem at all.

bkrabach:
**Correction to the line above** — one word was eaten by shell command substitution when I posted that comment (my own defect, third time in this batch; a quoted heredoc is the fix I keep failing to use first).

The bullet should read:

- Its own **`TestDiscoveryScopeParity`** caught the Windows failure that would otherwise have shipped as a vacuous PASS discovering 0 agents.

That guard is the reason this PR's problem was diagnosable rather than merged. It is preserved and strengthened in #365 — 13 tests now, of which 6 fail against this branch's v1.6.0 recipe, including two new ones pinning exactly what went wrong here: `test_discovery_survives_a_repo_path_containing_escape_sequences` and `test_discovery_fails_loudly_on_a_missing_repo_path`.

PR #365, opened 2026-09-06
fix(recipes): validate-agents classifies by `meta:`, not directory name -- and stops losing the repo path on Windows

bkrabach:
Builds on #364 (`lane/xe1u-validate-agents-discovery`, `038801f`), which is held DRAFT for exactly the two defects fixed here. **This PR contains that commit** — it is the mergeable unit; #364 should not merge alone.

Work item `model_performance-39z0`. **Draft on purpose — do not merge.**

## Headline

Both defects fixed. **The FAIL is gone.** The verdict does **not** return to bare `PASS` — it lands on **`⚠️ PASS WITH WARNINGS`**, and per the item that is the headline finding, named and **left unfixed**:

> `examples/agents/file-responder.md` — `NO_TOOLS_SECTION`. One WARNING, zero ERRORs.

Correct discovery surfaced it for the first time. It is a fixture demonstrating capability inheritance, so adding `tools:` would destroy what it demonstrates. **Reported, not fixed, not suppressed.** No threshold, severity, skip rule or exclusion set was touched.

## Before / after

Measured by executing the recipes' own **deterministic** steps — no LLM step, no API spend. `quality_classification.quality_level` **is** the verdict (`good→PASS`, `polish→PASS WITH SUGGESTIONS`, `needs_work→PASS WITH WARNINGS`, `critical→FAIL`); the report step renders that decision, it does not make it.

| | recipe | scanned | agents | non-agents | ERR | WARN | `quality_level` | verdict |
|---|---|---|---|---|---|---|---|---|
| baseline `5a9e07b` | v1.5.1 | — | 23 | — | 0 | 0 | `good` | **PASS** |
| before `038801f` | v1.6.0 | — | 32 | — | **4** | 1 | `critical` | **FAIL** |
| **after** | **v1.7.0** | **32** | **28** | **4** | **0** | 1 | `needs_work` | **⚠️ PASS WITH WARNINGS** |

**True agent count: 28** — measured, not assumed. 32 files match the scan; 4 declare no `meta:`.

**The harness is validated against the two paid runs it replaces**: it reproduces `run-c26b396f2f1e` (23/PASS) and `run-67388a546d75` (32/FAIL) exactly — same counts, same four `NO_FRONTMATTER` errors, same verdicts.

## (1) Classification by directory name

`context/agents/` merely collides on the word "agents". Its four files are context documents loaded via `context:` by `behaviors/agents.yaml:44-45` and `behaviors/tasks.yaml:18-19`; nothing spawns them, and they carry no frontmatter **because they are not agents**.

Discovery now classifies on **`meta:` presence — the loader's own contract** (`docs/AGENT_AUTHORING.md:5`: *"Agents ARE bundles … the only difference is the frontmatter key (`meta:` vs `bundle:`)"*).

A **classifier, not an exclusion**: `AGENT_SCAN_EXCLUDED_PARTS` untouched, no second list, all 32 candidates still walked, and every non-agent reported by name and reason in the new `non_agents_found` — the report and quick-approval path must now print a NON-AGENTS table. A real agent that lost its `meta:` block appears in that table instead of vanishing. Unparseable frontmatter still counts as an agent, so Phase 2 reports the YAML error.

## (2) The recipe found nothing on Windows

Every Python heredoc wrote `Path("{{repo_path}}")` — a path inside a **Python string literal**, where each backslash is an escape sequence. A GH Actions Windows checkout is `D:\a\<repo>\<repo>`, so `\a` parsed as `\x07`.

Reproduced on POSIX (`evidence/windows-path-repro.txt`) with a directory literally named `a\test\…`, running each version's own discovery body:

```
--- v1.6.0 (038801f) ---
exit code: 0
agents_found : []            <-- empty payload, EXIT 0
--- v1.7.0 (this branch) ---
exit code: 0
agents_found     : ['agents/probe.md']
non_agents_found : ['context/agents/notes.md']
```

`agents_found: []` **with exit 0** is the whole problem: structural validation then saw 0 agents / 0 errors, classification called that `good`, and the run reported **PASS having validated nothing** — strictly worse than the 23-of-32 under-count.

Fix: the path arrives via **`VALIDATE_AGENTS_REPO_PATH`** (no escape semantics) in all three affected steps — `environment-check`, `agent-discovery`, and `structural-validation`, whose `get_bundle_level_tools("{{repo_path}}")` would otherwise have produced false WARNINGs on Windows once discovery started working. Discovery also **exits non-zero** on a missing repo path, so no future path defect reaches a vacuous PASS quietly. A repo genuinely without agents is unchanged.

## The parity guard survived, and got stronger

It is what caught the Windows defect on Windows 3.11/3.12/3.13 while every POSIX leg stayed green. **Neither deleted nor relaxed** — 2 tests → 6:

- **scan parity** (`_agent_candidate_files`) is now pinned **separately** from agent parity, because a classifier that quietly narrowed the walk would still satisfy an agent-set comparison if the guard narrowed with it
- agent parity kept, plus a non-empty assertion
- `context/agents/*.md` are non-agents **and are named as such**
- the Windows defect pinned at its **cause** (no path in Python source)
- the Windows defect **reproduced on every platform**
- loud exit on a missing repo path

**6 of 7 fail against v1.6.0's recipe** (`evidence/parity-guards-FAIL-BEFORE.txt`); **13/13 pass here** (`evidence/parity-guards-PASS-AFTER.txt`).

## CI — all six legs green, including the three Windows ones

Run `34057350037`:

```
Tests (ubuntu-latest,  3.11/3.12/3.13)  pass
Tests (windows-latest, 3.11/3.12/3.13)  pass    <-- red on #364
```

**The Windows deliverable, proved on the real platform** by the same `TestDiscoveryScopeParity` that failed there with `assert set() == {...}` on #364.

## Tests

`uv sync --extra grpc-adapter && uv run pytest tests/ -q` → **1 failed, 1834 passed, 3 skipped**. The one local failure is `tests/test_sources.py::TestFileSourceHandler::test_resolve_existing_file` — environment-specific (a `/tmp` symlink `resolve()` collapses), **green in CI**, and `git diff origin/main -- tests/test_sources.py amplifier_foundation/sources` is empty. `recipes(operation="validate")`: `status: valid`, no warnings.

## Spend, and one honest gap

**$0.00 spent against a $0.00 authority** (`0 runs x 0 arms x $0 / 1.00`). No new paid `run-id` was purchased.

Executed at zero cost: three full deterministic replays (v1.5.1 / v1.6.0 / v1.7.0) over 32 candidates each — 12 deterministic step executions — plus the two-version Windows reproduction, the fail-before and pass-after guard runs, the full suite, and recipe schema validation; reproducing both prior run ids exactly.

The deliverable *"both recipes re-run … with run ids"* requires purchased runs while the authority sizing it is `0 runs x $0`. Those clauses cannot both hold — reported as a goal-authority defect rather than absorbed by overspending. The authority that would close it is `2 recipes x 1 run x <measured per-run $>`; **no prior lane recorded a dollar cost for these runs**, so that figure does not exist in evidence yet. Full accounting: `docs/lanes/39z0-agent-classifier-scope/DONE-NOTE.md`.

`validate-bundle-repo.yaml` is deliberately untouched (that is `model_performance-dfni`, sequenced after this because its count depends on this classifier), which also avoided its `bundle.dot`/`bundle.png` regen side effect entirely.

### DIFF
--- a/docs/lanes/39z0-agent-classifier-scope/DONE-NOTE.md
+++ b/docs/lanes/39z0-agent-classifier-scope/DONE-NOTE.md
@@ -1,0 +1,266 @@
+# Lane 39z0 — validate-agents: classify by `meta:`, not by directory name; and stop losing the repo path on Windows
+
+**Item:** `model_performance-39z0` (project `model_performance`)
+**Branch:** `lane/39z0-agent-classifier-scope`, branched from `lane/xe1u-validate-agents-discovery` (head `038801f`, draft PR #364)
+**Outcome:** **A — RESOLVED**, with one sub-deliverable recorded **NOT-POSSIBLE at the cap** (branch B applies to that clause only; see *Spend* below).
+**Recipe:** `recipes/validate-agents.yaml` v1.6.0 → **v1.7.0**
+
+---
+
+## Headline
+
+Both defects are fixed. **The FAIL is gone.** The verdict does **not** return to bare
+`PASS` — it lands on **`⚠️ PASS WITH WARNINGS`** — and per the item's own instruction that
+is the headline finding, named and **left unfixed**:
+
+> `examples/agents/file-responder.md` — `NO_TOOLS_SECTION` (no explicit `tools:` section,
+> relies on inheritance). One WARNING, zero ERRORs.
+
+This is a **genuinely new finding surfaced by correct discovery**, and it is exactly the
+file the item's KNOWN section flags as *"a fixture demonstrating capability inheritance"*.
+Adding a `tools:` block to it would destroy what the fixture demonstrates, so it is
+**reported, not fixed here, and not suppressed**. No threshold, severity, skip rule or
+exclusion set was touched to soften it.
+
+---
+
+## The three measurements
+
+All numbers below come from executing the recipes' own **deterministic** steps
+(`environment-check → agent-discovery → structural-validation → quality-classification`)
+verbatim through bash, at **$0.00 API spend** — see *Spend* for why, and
+`replay_deterministic_phases.py` for the harness. `quality_classification.quality_level`
+**is** the verdict: the recipe's own Verdict Selection maps
+`good → PASS`, `polish → PASS WITH SUGGESTIONS`, `needs_work → PASS WITH WARNINGS`,
+`critical → FAIL`. The LLM report step renders that decision; it does not make it.
+
+| | recipe | scanned | agents | non-agents | ERR | WARN | `quality_level` | **verdict** |
+|---|---|---|---|---|---|---|---|---|
+| baseline (`5a9e07b`, pre-xe1u) | v1.5.1 | — | **23** | — | 0 | 0 | `good` | **PASS** |
+| before (`038801f`, xe1u head) | v1.6.0 | — | **32** | — | **4** | 1 | `critical` | **FAIL** |
+| **after (this branch)** | **v1.7.0** | **32** | **28** | **4** | **0** | 1 | `needs_work` | **⚠️ PASS WITH WARNINGS** |
+
+**The true agent count is 28.** Measured, not assumed: 32 files match the scan, 4 of them
+declare no `meta:` and are not agents.
+
+**The replay harness is validated against xe1u's two paid runs.** It reproduces
+`run-c26b396f2f1e` (23 agents / PASS) and `run-67388a546d75` (32 agents / FAIL) exactly —
+the same counts, the same four `NO_FRONTMATTER` errors, the same verdicts. That agreement
+is what licenses the third row as a real measurement rather than an argument.
+
+Evidence: `evidence/deterministic-before-after.txt`,
+`evidence/deterministic-baseline-v1.5.1.txt`.
+
+---
+
+## Defect 1 — classification by DIRECTORY NAME
+
+v1.6.0 treated every `agents/*.md` as an agent. `context/agents/` merely **collides on the
+word "agents"**: its four files are context documents loaded via `context:` by
+`behaviors/agents.yaml:44-45` and `behaviors/tasks.yaml:18-19`. Nothing spawns them, and
+they carry no frontmatter **because they are not agents** — so they produced four
+`NO_FRONTMATTER` ERRORs and flipped the verdict on a repo with zero real agent defects.
+
+**Fix: classify on `meta:` presence — the loader's own contract**, quoted from this repo's
+`docs/AGENT_AUTHORING.md:5`:
+
+> "Agents ARE bundles. They use the same file format and are loaded via `load_bundle()`.
+> The only difference is the frontmatter key (`meta:` vs `bundle:`)."
+
+Three properties this fix deliberately has:
+
+1. **It is a classifier, not an exclusion.** `AGENT_SCAN_EXCLUDED_PARTS` is untouched, no
+   second exclusion list exists, and **every candidate is still walked**. `candidates_scanned`
+   is still 32.
+2. **Nothing is dropped silently.** Every non-agent is reported by name and reason in the
+   new `non_agents_found` / `non_agent_reasons`, and both the report and the quick-approval
+   path are now *required* to print a NON-AGENTS table. A classifier that quietly discarded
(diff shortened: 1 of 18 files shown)


bkrabach:
## Manager verification — all 6 gates re-run by me. Merging the combined result (xe1u's widening + 39z0's classifier).

Head `9c4e53c`, base `5a9e07b`. **CI 7/7 SUCCESS — including Windows 3.11/3.12/3.13, where #364 was red.**

| gate | decisive fact |
|---|---|
| **1. classifier** | I ran the predicate myself: **candidates_scanned 32 → agents 28, non_agents 4**. The 4 excluded are exactly `context/agents/{delegation-instructions,multi-agent-patterns,session-repair-knowledge,session-storage-knowledge}.md` |
| **2. verdict** | **0 errors** (was 4), **1 warning**, at the true count 28 |
| **3. Windows** | root cause reproduced by me — see below. Discovery `sys.exit(1)` on a missing path (`:489`), so **no vacuous PASS is reachable** |
| **4. parity guard** | **13 passed** on branch; **6 fail against v1.6.0** — strictly stronger, not relaxed. `AGENT_SCAN_EXCLUDED_PARTS` still the single list |
| **5. not weakened** | zero threshold/severity/skip changes; the only diff lines matching are comments *stating* that nothing was changed |
| **6. suite** | **1955 passed, 2 failed** — both the known pre-existing pair I verified reproduces at `origin/main` |

### The 23 → 28 breakdown, measured not assumed

Previously walked (`agents/` + `bundles/*/agents/`) = **23**. The **5 newly counted agents**:

```
examples/agents/file-responder.md
experiments/build-up/agents/coder.md
experiments/build-up/agents/explorer.md
experiments/build-up/agents/planner.md
experiments/build-up/agents/tester.md
```

23 + 5 = 28. The four `context/agents/*.md` are **candidates that are correctly classified out**, not files hidden from the walk — all 32 are still scanned, and every exclusion is reported by name and reason in the new `non_agents_found`.

### The Windows defect was never a path-separator problem — and that finding is the best thing here

I reproduced F2 myself:

```
interpolated-into-source yields: 'D:\x07mplifier\test'   <- \a became BEL
via environment yields         : 'D:\\amplifier\\test'   <- intact
```

v1.6.0 interpolated the repo path **into Python source**, where `\a` is an escape sequence. The path was corrupted before any globbing happened, so discovery returned `agents_found: []` **with EXIT 0** — a valid empty payload that reads as success. Not a glob bug, not `os.sep`. The fix routes the path through `VALIDATE_AGENTS_REPO_PATH` in the environment, in **all three** affected steps — including `structural-validation` (F3), which carried the identical defect and was invisible only because discovery already returned nothing. It would have produced false WARNINGs the moment discovery started working.

**Both new behaviours are pinned by tests that fail against v1.6.0**: `test_discovery_survives_a_repo_path_containing_escape_sequences` and `test_discovery_fails_loudly_on_a_missing_repo_path`.

### It did NOT return to a bare PASS, and said so as the headline

`examples/agents/file-responder.md` raises `NO_TOOLS_SECTION` → `needs_work` → **PASS WITH WARNINGS**. It is a fixture demonstrating capability inheritance, so **adding `tools:` would destroy what it demonstrates**. Reported, left unfixed, nothing suppressed — exactly what the goal demanded of a newly-surfaced finding. I asked for PASS at 28; PASS-with-a-named-warning is the honest version of that, and hiding the warning to reach bare PASS would have been the failure.

**F4 is the self-aware part**: a classifier introduces its own silent-drop risk — a real agent that lost its `meta:` block would simply stop being validated. Mitigated structurally by requiring the NON-AGENTS table in the report, so an exclusion can never be silent.

**F5 — a goal-authority defect against me, and it is right.** A `$0.00` authority cannot fund a deliverable requiring paid recipe runs. It reproduced both prior paid runs (`run-c26b396f2f1e` 23/PASS, `run-67388a546d75` 32/FAIL) **deterministically with zero API spend** by executing the recipes' own phases, then recorded the new-run-id as NOT-POSSIBLE at the cap rather than quietly skipping it. Second lane this batch to audit my arithmetic.

PR #366, opened 2026-09-06
fix(recipes): validate-bundle-repo reported "0 errors" because it could not see the error class

bkrabach:
## The count was the symptom. The validator could not fail.

`recipes/validate-bundle-repo.yaml`'s `agent-description-validation` reported
`agents_checked: 32, errors: []` on this repo. That figure was quoted last cycle as evidence
that the repo's agents were clean. **On this class of problem it could not have been anything
else.**

`extract_description()` returned an **empty string** for six different failures — unreadable
file, absent frontmatter, unparseable YAML, non-mapping frontmatter, non-mapping `meta:`,
absent `description:` — and for a genuinely empty description too. `""` then passed every
check: 0 tokens is under budget, and it contains no `<example>` and no `<commentary>`. So a
**real agent whose frontmatter was missing or malformed was reported as checked and clean.**

### Measured, on a fixture where four of five "agents" are broken

```
BEFORE  v3.12.0 :  agents_checked: 5,  errors: 0,  passed: True
AFTER   v3.13.0 :  agents_checked: 4,  errors: 3,  passed: False
                   [agent_frontmatter_invalid] agents/broken-yaml.md      (unparseable_frontmatter)
                   [agent_description_missing] agents/no-description.md   (no_description_key)
                   [agent_description_missing] agents/empty-description.md (description_empty)
                   NON-AGENT                   agents/no-frontmatter.md   (no_frontmatter)
```

### And on this repo — the count deliverable

| | recipe | candidates | `agents_checked` | non-agents | ERR |
|---|---|---|---|---|---|
| before (`e495755`) | v3.12.0 | not reported | **32** | not reported | 0 |
| **after** | **v3.13.0** | **32** | **28** | **4** | **0** |

`errors: []` is preserved. The four excluded files are `context/agents/*.md` — context
documents loaded via `context:`, named individually with reason `no_frontmatter`.

All figures come from executing the recipe's own step body verbatim, at **$0.00 API spend**
(`docs/lanes/dfni-validate-bundle-repo-count/measure.py`).

## What changed

1. **`extract_description()` returns `(status, description)`.** `"ok"` only when a non-empty
   string description was actually read. Everything else is an ERROR —
   `agent_frontmatter_invalid` (the file is broken) or `agent_description_missing` (no usable
   description), each carrying the precise `reason`. A non-ok file never reaches the budget or
   example checks; its `description_tokens` is `null`, not a passing `0`.
2. **The classifier from `validate-agents` v1.7.0, copied BYTE-FOR-BYTE.** Files under
   `agents/` are classified on `meta:` presence — the loader's own contract
   (`docs/AGENT_AUTHORING.md:5`) — not on the directory name. Not an exclusion:
   `EXCLUDED_DIRS` is untouched, every candidate is still walked, and every non-agent is
   reported by name and reason in `non_agents_found` (39z0 F4 preserved; the report now
   *requires* a NON-AGENTS table).
3. **The Windows defect, in all TEN Python steps.** Every one wrote
   `repo_path = "{{repo_path}}"` — a path inside a Python string literal, where `\a` in
   `D:\a\<repo>\<repo>` parses as `\x07`. The path did not exist, the walk found nothing, and
   the step reported a clean result over **zero files**. All ten now read
   `$VALIDATE_BUNDLE_REPO_PATH`. This recipe is worse-affected than `validate-agents` was:
   every step carries `on_error: continue`, so the empty walk read as success.
4. **The same pattern, cross-checked across every extractor in the recipe.** Two more
   instances found and fixed, three confirmed safe:
   - `behavior-reference-hygiene`'s `extract_frontmatter_yaml()` returned `{}` for every
     failure, and `{}` satisfies both of that step's checks vacuously → `behavior_frontmatter_invalid`.
   - `mode-validation`'s `extract_frontmatter()` didn't falsely pass — it silently **dropped**.
     A mode with broken frontmatter stopped being validated with nothing in the output to say
     so → `mode_frontmatter_invalid` + `non_modes_found`.
   - `parse_yaml_bundle()`, `parse_yaml_section()`, `get_bundle_details()` — safe, unchanged,
     reasons recorded in the DONE-NOTE.
5. **A dead error channel, closed.** `quality-classification` read
   `behavior_reference_hygiene.warnings` but never its `errors`. Nothing was ever lost (the
   step emitted none), but an unread error channel is this PR's whole subject in a different
   place.

## Single-sourcing: how "the same classifier, not a copy" was achieved

Recipe steps are self-contained heredocs with **no import mechanism** between two `.yaml`
recipes, and a Python helper in `amplifier_foundation/` is explicitly allowed to be
un-importable in the third-party repos these recipes run against — so making *scope* depend on
it would trade a visible under-count for an invisible one. `xe1u` declined that route for the
same reason; I reached the same conclusion and took the substitute this item names, then made
it stronger:

**The classifier block is byte-identical across both recipes, and a test compares them
character for character.** Textual identity is pinned *in addition to* behavioural parity, not
instead of it — the scan set and the classified set are both compared against the guard, kept
separate so a classifier that quietly narrowed the walk cannot hide behind a narrowed guard.

## Tests — 9 new guards, `TestDiscoveryScopeParity` 7 → 16 (nothing deleted, nothing relaxed)

- Against `origin/main`'s recipe: **9 failed, 7 passed**
- Against this branch: **16 passed**

Every guard runs the recipe's **real step body**, never a re-implementation.

## Suite

`uv run pytest tests/ -q` → **1 failed, 1843 passed, 3 skipped**. The one failure is the known
pre-existing `test_sources.py::TestFileSourceHandler::test_resolve_existing_file`;
`git diff origin/main -- tests/test_sources.py amplifier_foundation/sources` is empty.

`recipes(operation="validate")`: **`status: valid`, no warnings.**
`git status` is clean of `bundle.dot`/`bundle.png` — the full recipe was never executed, so
the `6phe` F1 regen side effect never fired.

Three tests in `tests/test_module_dep_resolvability_check.py` were updated because their
harness interpolated `{{repo_path}}` into the step source — the exact defect being removed. No
assertion was weakened.

## Spend

**$0.00.** Zero API spend, zero DTU, zero infrastructure. The item's stated preference —
reproduce deterministically rather than buy a run — was sufficient for every deliverable, so
nothing is NOT-POSSIBLE and there is no unspendable residue. A paid run carrying a `run-id`
was deliberately not purchased; 39z0's open note (no lane has yet recorded a measured per-run
dollar cost for these recipes) is left open and named.

Full detail, evidence and deviations:
`docs/lanes/dfni-validate-bundle-repo-count/DONE-NOTE.md`

---
Generated with Amplifier


bkrabach:
## Manager verification — **NOT merging.** Gates 1-4 hold; **gate 5 fails: Windows CI is red on this PR's own new test.**

Head `ffcb08e`, base `e495755`. Ubuntu 3.11/3.12/3.13 **green**; **Windows 3.11/3.12/3.13 RED**.

| gate | result |
|---|---|
| **1. count 32 → 28** | ✅ reproduced; `errors: []` preserved; `candidates_scanned: 32`, `non_agent_count: 4` — the same four `context/agents/*.md` files `39z0` classified out |
| **2. single classifier** | ✅ no second implementation — byte-identical to `validate-agents` v1.7.0's, pinned by `test_both_recipes_carry_the_same_classifier_source` |
| **3. `extract_description()`** | ✅ genuinely fixed, not filtered upstream — returns `(status, description)`; the synthetic fixture goes **5 checked / 0 errors / PASS → 4 checked / 3 errors / FAIL** |
| **4. parity guard** | ✅ **22 tests, all passing** on POSIX; neither deleted nor weakened |
| **5. CI all platforms** | ❌ **Windows red** |

### The failing gate, with the cause

```
tests/test_anchors_bundles_dry.py:882
    assert [error["file"] for error in broken] == ["modes/broken.md"]
E   AssertionError: a mode with unparseable frontmatter was skipped silently:
    {'phase': 'mode_validation', 'modes_checked': 1, 'candidates_scanned': 3, ...}
```

**This is not the BEL/source-interpolation class `39z0` fixed** — discovery works here (`candidates_scanned == 3` and `modes_checked == 1` both passed, and the env-var routing is correctly applied to all ten steps). It is a **different, real cross-platform defect one layer further out**:

```
recipes/validate-bundle-repo.yaml:3431   rel_path = str(mode_file.relative_to(path))
recipes/validate-bundle-repo.yaml:3416   mentions.append(str(md_file.relative_to(path)))
recipes/validate-bundle-repo.yaml:955    f"Failed to parse {mp.relative_to(path)}: {err}"
```

`str()` on a `Path` yields **OS-native separators**, so on Windows the payload reports `modes\broken.md` where every consumer — including this PR's own tests and the `non_modes_found` assertion two lines below — expects `modes/broken.md`. **4 sites** carry this shape.

**Worth fixing rather than adjusting the test**, for the same reason the lane's own F5 exists: a validator's machine-readable output should be **stable across platforms**. Any downstream matcher on a reported file identifier silently stops matching on Windows — the identical failure mode to a mode being "silently dropped", just relocated to the consumer. The one-line remedy is `.relative_to(path).as_posix()` at each site.

### What this PR gets right, and it is substantial

The headline is correct and larger than the item I filed: **`extract_description()` collapsed six distinct failure conditions and a genuinely-empty description into one empty string**, and an empty string satisfied every downstream check (0 tokens is under budget; an empty string contains no `<example>` and no `<commentary>`). A malformed-frontmatter agent was reported **checked and clean**.

It also **took the optional cross-check and found the pattern was three defects, not one**:
- **F4** — `behavior-reference-hygiene`'s `extract_frontmatter_yaml()` returned a bare empty dict on every failure, and an empty dict satisfies both of that step's checks **vacuously**;
- **F5** — `mode-validation` did not falsely pass, it **silently dropped**: a mode with broken frontmatter stopped being validated with nothing in the output saying so. Fixed with an ERROR plus `non_modes_found` naming every excluded file — **`39z0`'s F4 mitigation, correctly generalised to modes**;
- **F6** — a **dead error channel**: `quality-classification` read `behavior_reference_hygiene.warnings` but never its `errors`. Nothing was lost yet because that step emitted none — but an unread error channel is one append away from losing a real finding.

And **F3 is the part I want on record**: the repo-path defect was in **all ten** Python steps of this recipe, not just the one my item named — and it is **worse here than in `validate-agents`**, because every step carries `on_error: continue`, so a path that silently resolved to nothing produced a **clean payload** rather than a visible failure.

Fix the four `str(relative_to)` sites, get Windows green, and this merges. Nothing else is outstanding.

bkrabach:
## Gate 5 cleared — Windows now green 6/6. Merging.

I fixed the blocker myself rather than sending it back, on the same precedent as #310/#313/#306: five sites emitting OS-native path separators, one line each.

```
recipes/validate-bundle-repo.yaml:955   f"Failed to parse {mp.relative_to(path)}: {err}"
                                :3416   mentions.append(str(md_file.relative_to(path)))
                                :3431   rel_path = str(mode_file.relative_to(path))
                                :4167   doc_pip_hits.append(str(doc_file.relative_to(path)))
                                :4303   rel_path = str(file_path.relative_to(path))
```

All five now use `.relative_to(path).as_posix()`. **Fixed at the source rather than by relaxing the assertions**, for the reason this PR's own F5 rests on: a validator's machine-readable output should be **stable across platforms**. A downstream matcher on a reported file identifier silently stops matching on Windows — the identical failure mode to a mode being "silently dropped", relocated to the consumer.

`4b45b45` — POSIX unchanged (22 passed before and after), recipe still parses at 3.13.0, and **Windows 3.11/3.12/3.13 went red → green**.

Final gate table:

| gate | result |
|---|---|
| 1. count 32 → 28 | ✅ `errors: []` preserved; `candidates_scanned: 32`, `non_agent_count: 4` — the same four `context/agents/*.md` |
| 2. single classifier | ✅ byte-identical to `validate-agents` v1.7.0's, pinned by `test_both_recipes_carry_the_same_classifier_source` |
| 3. `extract_description()` | ✅ genuinely fixed — fixture goes **5 checked / 0 errors / PASS → 4 checked / 3 errors / FAIL** |
| 4. parity guard | ✅ **22 tests**, neither deleted nor weakened |
| 5. CI all platforms | ✅ **6/6 green** |

This completes the validator-integrity arc: **6phe** (content) → **xe1u** (measurement; not merged, but it exposed both defects) → **39z0** (classifier + the BEL source-interpolation bug) → **dfni** (the extractor, the count, and the pattern across three more steps). Both recipes now report the same population, **28**, by construction.

bkrabach:
## Gate 5 cleared — Windows green. Merging.

I fixed the blocker myself rather than leave verified-good work stranded, same precedent as #310 (Windows/bash) and #313 (macOS `stat -c %Y`): the gate was small, mechanical, and not a defect in the lane's reasoning.

**Five sites** reported file identifiers via `str(Path.relative_to(...))`, which yields **OS-native separators** — so on Windows the payload said `modes\broken.md` where every consumer, including this PR's own assertions, expects `modes/broken.md`. Rewrote each to `.as_posix()` (`4b45b45`).

**Fixed rather than relaxing the assertions**, for exactly the reason this PR's own F5 rests on: a validator's machine-readable output should be **stable across platforms**. A downstream matcher on a reported file identifier silently stops matching on Windows — the identical failure mode to a mode being 'silently dropped', relocated to the consumer.

POSIX unchanged: **22 passed** before and after; recipe still parses at 3.13.0. CI now **7/7**, Windows 3.11/3.12/3.13 included.

Final gate table:

| gate | result |
|---|---|
| 1. count 32 → 28 | ✅ reproduced, `errors: []` preserved, same four `context/agents/*.md` classified out |
| 2. single classifier | ✅ byte-identical to validate-agents v1.7.0, pinned by `test_both_recipes_carry_the_same_classifier_source` |
| 3. `extract_description()` | ✅ genuinely fixed: fixture goes 5 checked/0 errors/PASS → 4 checked/3 errors/FAIL |
| 4. parity guard | ✅ 22 tests, none deleted or weakened |
| 5. CI all platforms | ✅ after `4b45b45` |

The validator-integrity arc closes here: **6phe** (content) → **xe1u** (measurement, closed superseded) → **39z0** (classifier + the BEL source-interpolation defect) → **dfni** (the extractor that could not fail, plus the same pattern in two more steps and a dead error channel). Both recipes now report the same population, **28**.

PR #372, opened 2026-09-07
perf(head): ship the lean head -- delegate preamble + anchors context, behind an Anthropic guardrail shown RED then GREEN

bkrabach:
Ships the **foundation-owned share** of the lean head from `model_performance-zc6t`, behind an Anthropic non-regression guardrail that was **shown red before it was allowed green**.

## Why this ships without a new measurement

`model_performance-g7h3` bought 98 end-to-end claude-opus-5 / S7-17 runs (49/arm, $428.10) and **all three pre-registered estimators excluded zero — the first time in this program**:

| estimator | n | effect | 95% CI |
|---|---|---|---|
| primary (VALID only, pre-registered) | 11 / 8 | **−13.57 %** | [−22.27 %, −4.86 %] |
| sensitivity (ALL runs) | 49 / 49 | −16.42 % | [−23.29 %, −9.56 %] |
| block-paired (time-trend robust) | 7 pairs | −14.44 % | [−26.83 %, −2.06 %] |

Co-primary quality is **cited from `model_performance-5zp`**: one-sided 95 % LB **−7.15 pp** against the frozen **−10 pp** margin — clears. `B`, the behavioural term three lanes called "the whole remaining question", is **~zero (−0.33 %)** against the **+16.6 %** a wash would have required.

**Spend for this PR: $0.** No `$/task` re-measurement was needed, authorised, or bought.

> `B` has the **opposite sign on the two providers** (−10.7 % requests on opus-5 vs +24.5 % on terra in `3d2`). It is measured, not understood, and must never be extrapolated across a provider. **Terra remains HOLD-because-underpowered (MDE 21.7 %) — which is not a measured wash.**

## What changed

| Artifact | stock | lean | saved |
|---|---:|---:|---:|
| `delegate` **preamble** (`modules/tool-delegate`) | 1,190 | **938** | 252 |
| `bundles/anchors/context/system.md` | 1,358 | **1,142** | 216 |
| `bundles/anchors-amp-dev/context/amplifier-ecosystem.md` | 1,432 | **1,300** | 132 |
| **total** | **3,980** | **3,380** | **600** |

The `delegate` preamble is **byte-identical to the v1 text** measured on the wire in all 98 runs — pinned against a vendored copy of `probes/bji-lean-head/v1_tools.json`, not against a paraphrase of it.

**Two notes the census reader needs:**

* **`delegate` is the preamble only.** The agent catalog is appended by the tool at runtime and is out of scope (stock 12,904 / v1 3,597 — a 9,307-char difference **not** included above).
* **The last two files are one v1 span (#9).** It was authored against `anchors-amp-dev/context/system.md`, which `ux32`'s DRY refactor split in two, so the compression is applied across **both** halves.

## Fidelity beats compression — and here it cost bytes

Every rule, constraint, command and pointer in the stock text of **all 23 targets** was extracted and checked against the lean text. Four flags; three false positives (a collapsed operation list, a Unicode ellipsis, an invented example filename) and one real weakening in a description this repo does not own. Working shown in `docs/lanes/zc6t-lean-head-ship/FINDINGS.md`.

Today's ecosystem file carries three things the v1 span never saw — the `context-intelligence:session-navigator` pointer, the `graph-analyst` fallback, and the `python scripts/amplifier-session.py {diagnose,repair,rewind,info,find}` command with its *"there is no `amplifier session repair` subcommand"* correction. **All three are restored, at +450 chars against a v1-faithful rewrite**, and pinned by `TestFidelityBeatsCompression`. A saving bought by deleting a real instruction is a behaviour change with nothing pointing back at the commit that caused it.

## The guardrail failed before it passed

| | SHA | CI |
|---|---|---|
| **RED** — guardrail, stock head | `5833f84f843912f2e5bb696225a4d63fafc98256` | [run 34143330945](https://github.com/microsoft/amplifier-foundation/actions/runs/34143330945) — **4 failed**, 1869 passed |
| **GREEN** — guardrail, lean head | `46fbd454c99e90662aee3447efff7d23115969c0` | [run 34143774767](https://github.com/microsoft/amplifier-foundation/actions/runs/34143774767) — **success**, all 6 legs |

Exactly four failures in the red run, all of them the guardrail, on all six OS/Python legs. Nothing else in the suite moved.

**What each assertion can and cannot prove** — stated in the test module, because a guardrail whose limits are not written down gets read as proving more than it does:

* **(a) head census** — fully proven. Exact char counts of real sources. This is the assertion that goes red.
* **(b) one cold head write per session** — the **precondition** is proven (the owned head text is deterministic: no timestamp, absolute path, sha or session id, and it composes identically twice). The `cache_read` behaviour itself is **not** asserted — a static test cannot observe it — and stays measured in `head_stability.py` and the 98 g7h3 runs.
* **(c) non-inferiority** — a **citation pin**, not a measurement. Fails if the frozen margin is loosened or the cited lower bound is edited to one that no longer clears it.

## Measured on the wire, $0

The census hook dumps the fully-composed Anthropic `params` and calls `os._exit()` **before** the HTTP call — a genuine wire head, no tokens purchased.

```
bundle anchors-amp-dev · claude-opus-5 · 2026-09-07

              instr      tools       head    context_file blocks
BEFORE       99,557    220,853    320,410           24
AFTER        99,209    220,587    319,796           24
delta          -348       -266       -614
```

`"You are Amplifier"` stays at char **247**; the block count does not move.

> **The item's 48,249-char threshold does not transfer to this host.** This head measures **320,410 chars — 3.8× that** — because it composes 86 tools and 24 context files where the eval container composed 14 tools. 48,249 describes a *bundle composition*, not the product. **That is why guardrail (a) pins per-artifact budgets rather than one whole-head absolute**: a per-artifact budget is portable across bundle sets and still fails on exactly the regression the absolute was meant to catch.

## Also fixed

The stock `delegate` preamble emitted the self-delegation line **twice** — once hardcoded in `base_description`, again from the feature registry. **76 chars on every request of every session, on both wires.** Removed as a side effect of folding `"self"` into the single `- agent:` bullet.

## Scope excluded, deliberately

* **`web_search` and `mode` are already byte-identical to their v1 text** (30 and 198 chars) and are correctly left untouched. The item flagged `web_search`; it did **not** flag `mode` — so **11 tool descriptions actually change, not 13**.
* **v1 spans 10 and 11 are hook-generated** (`<system-reminder source="routing-matrix">`, `<system-reminder source="hooks-skills-visibility">`). No file produces them; a renderer emits them per turn. **Not hand-edited** — that would be editing a symptom. Their lean blocks would be 861 + 5,350 = **6,211 chars**, which is part of why the realised saving is short of the shim's 48,249. A follow-up item covers the renderer.
* **21 of the item's 23 targets live in 13 repositories this lane does not contain** (the 23 span 14 repos in total). Per `GOAL.md` — *"Never touch other repos"*, and its explicit clause that an unreachable deliverable is *"a DEFECT IN THIS GOAL, not a task … report it, ship the patch as an artifact, and resolve"* — the 19 that need a change ship as drop-in patch artifacts (`docs/lanes/zc6t-lean-head-ship/patches/`, **11,738 chars of saving**: 10 tool descriptions + 9 context files) and the defect is reported as **F1** in `FINDINGS.md`. The constraint is the prohibition, not availability: 2 of those repos are present on this host and were still not touched.

## Tests

`1,872 passed` locally; `1,869 passed` on CI. The only local failure is the known pre-existing `tests/test_sources.py::TestFileSourceHandler::test_resolve_existing_file`, confirmed failing on clean `HEAD` and passing on CI (it is host-dependent). `190 passed` in `modules/tool-delegate/tests/`.

`test_delegate_self_delegation_disabled.py`'s `FEATURE_LINE` literal was **updated, not deleted** — `agent="self"` moved into the `- agent:` bullet. The invariant is unchanged, and a second **literal-independent** test now asserts it so a future rewording cannot re-open the gap.

**Default-mode byte identity does not hold, by design** — this change exists to move head bytes. What is verified is that nothing *else* moved.

---

Refs `model_performance-zc6t`, `model_performance-g7h3`, `model_performance-5zp`, `model_performance-skw3`, `model_performance-3d2`.

Generated with Amplifier


bkrabach:
## Manager verification — guardrail proven red-then-green. Merging.

Head `e49b0389`. **CI 7/7 SUCCESS** (ubuntu + Windows × 3.11/3.12/3.13 + cla). **Spend $0.00 of $15.**

### The guardrail was shown RED before it was allowed green — I verified both runs resolve

| run | sha | conclusion | detail |
|---|---|---|---|
| [34143330945](https://github.com/microsoft/amplifier-foundation/actions/runs/34143330945) | `5833f84f` | **failure** | 4 failed, 1869 passed — **every failure the guardrail, on all 6 OS/Python legs** |
| [34143774767](https://github.com/microsoft/amplifier-foundation/actions/runs/34143774767) | `46fbd454` | **success** | all 6 legs |
| [34143961273](https://github.com/microsoft/amplifier-foundation/actions/runs/34143961273) | `e49b0389` | **success** | head |

I confirmed the red run's failures landed in the **Tests** jobs, not in setup or lint by accident. **A guardrail that has never failed is decoration**; this one has now failed for the right reason.

### What actually shipped here, and what did not

**In this repo:** the `delegate` preamble **1,190 → 938 chars**, byte-identical to the v1 text measured on the wire in all 98 `g7h3` runs; span 9 across **both** halves of the file `ux32` split (`anchors/context/system.md` 1,358 → 1,142 and `anchors-amp-dev/context/amplifier-ecosystem.md` 1,432 → 1,300); and a pin test (`tests/test_lean_head_guardrail.py`) against a vendored verbatim v1 fixture so it cannot drift back.

**21 of 23 targets live in 13 other repositories this lane does not contain**, so they ship as **drop-in patch artifacts** under `docs/lanes/zc6t-lean-head-ship/patches/` — **3,926 chars** of tool-description saving and **7,812 chars** of context-file saving, each diffed and fidelity-checked. **That is my goal's defect, not the lane's**: I wrote "upstream order — foundation, then modules, then bundles, then app-cli" and then launched the lane with a single foundation checkout. It reported the contradiction (F1) instead of quietly editing outside its repo, which is the correct call.

### Four findings worth more than the bytes

**F7 — the stock delegate preamble emitted the self-delegation line TWICE**, hardcoded *and* from the feature registry: **76 chars on every request of every session, on both wires**, for as long as it has existed. Fixed here.

**F3 — `mode` is byte-identical to its v1 text (198 chars) and my goal did not flag it**, so **11 tool descriptions actually change, not 13**. `web_search` was the one I *did* flag as a likely no-op, and it verified as one (30 chars). The lane checked both rather than editing to match my count.

**F5 — the item's 48,249-char threshold does not transfer.** It describes the eval container's composition (14 tools); this host's `anchors-amp-dev` head measures **320,410 chars** (86 tools, 24 context files). So guardrail (a) pins **per-artifact budgets**, which are portable, instead of a whole-head number that would be meaningless off one machine. Head census here: **320,410 → 319,796 (−614)**, owned artifacts **3,980 → 3,380 (−600)**, context-file blocks **24 → 24**.

**F6 — a near-miss on a shared host, self-reported:** `amplifier source add` **ignores `AMPLIFIER_HOME`** and writes to the real `~/.amplifier/settings.yaml` (while reporting *"Scope: global"*). Caught from its own output, reverted within a minute, verified clean. Worth filing against the CLI.

### Scope note

The two hook-generated `<system-reminder>` blocks (**861 + 5,350 chars**) are **excluded by the goal's own instruction** — no source file produces them, and hand-editing captured output would be editing a symptom. They are `model_performance-z6wa`. Their share of the shim's 48,249 is unreachable until that renderer change lands, and the PR body says so rather than letting the shortfall look like a failed rewrite.

PR #373, opened 2026-09-07
feat(validators+docs): move the description rules from convention to enforcement

bkrabach:
## The problem

PR #341 set the no-`<example>` policy for description surfaces in
`context/shared/description-authoring-principles.md`. It was applied inside
`amplifier-foundation` and **nowhere else**, because nothing at authoring or
validation time said otherwise. A sweep a year later found **29 files across 6
repos** still shipping example blocks in descriptions and fixed them **by
hand** (android-tester 8/8 → 0, browser-tester 6 → 0, reality-check 10/10 → 0,
dot-graph 28/28 → 0 with its agent descriptions 14,615 → 6,484 chars,
context-intelligence 3 → 0, plus two forks).

A rule that lives only as a convention is a rule that quietly stops being true.
This PR moves it to where it executes.

## What changed

**Validators** — `validate-bundle-repo` v3.14.0 → **v3.15.0**, `validate-agents`
v1.7.0 → **v1.8.0**.

1. **The length gate is CHARACTERS, and it can now fire.** Both recipes gated on
   tokens (WARN >300, ERROR >600 = 1,200 / 2,400 chars at chars/4). Measured
   across 9 bundle repos, that ERROR line sat **above the longest description in
   every already-aligned repo** (foundation: n=24, mean 410, max 761), so on an
   aligned repo it was unfireable. Now WARN >600 chars, ERROR >1,200. Foundation:
   5 warnings, **0 errors**. The unswept corpus: 17 of 54 agents over the ERROR
   line. The two numbers are identical in both recipes and a test fails the build
   if they diverge.
2. **NEW Phase 2.82 — skill descriptions.** The example policy was written for
   every description surface and enforced on agents only. Skill frontmatter
   renders into the `hooks-skills-visibility` block on every request exactly as
   an agent description renders into the delegate catalog. Cap 400 / ERROR 800,
   plus v3.13.0's lesson applied from the start: every way of failing to *read* a
   description gets its own status and its own ERROR.
3. **NEW Phase 2.84 — awareness redundancy.** Two measurable rules, **stated in
   the step** so a finding can be audited rather than trusted. Calibrated on 36
   real context files: R1 fires 0 times (max coverage observed 0.333), R2 fires 3
   — each genuinely when-to-use prose plus a pointer. `best_coverage` is reported
   for every file either way.
4. **NEW Phase 2.86 — bundle head cost**, the number a bundle author never sees.
   WARNING at 4,000 chars, **justified against the measured bundles**: the two
   engineered for head cost measure 2,483 and 1,760; the nearest unengineered one
   5,479, then 14,198 / 15,002 / 26,368 / 29,684 / 87,555. An unresolvable
   include is not charged a flat guess — the total is marked `is_lower_bound` and
   the refs are named.
5. **Fail-before, reproducible by anyone.** `test-fixtures/misaligned-bundle/`
   violates all four at once; the tests run the recipes' **own step bodies**
   against it. Point `ALIGNMENT_RECIPE_DIR` at the pre-change recipes:

   | | result |
   |---|---|
   | origin/main **4384805** | **23 failed, 1 passed** |
   | this branch | **24 passed** |

   The one test that passes at main is the one that should — a 599-char clean
   description is clean under both gates.

**Docs** — the rules now state they apply to **every** bundle, with the bill
measured rather than asserted: the runtime agent catalog at **12,904 → 3,597
chars** on the wire, `g7h3`'s **−13.57% $/task, CI [−22.27%, −4.86%]**, all three
pre-registered estimators excluding zero. The lean head's 84,319 → 48,249 is
cited **as a bundle composition, not as a universal figure** — the same commit
measured 320,410 chars of head on a host composing 86 tools. New V7 (shape),
new BUNDLE_GUIDE sections for awareness and for refreshing.

**Refresh path** — `recipes/refresh-descriptions.yaml`. It does not carry a copy
of the rules (it executes `validate-bundle-repo`'s own step bodies and fails loud
if it cannot find them), and it does not let a shortening pass drop a routing
fact silently: the proposal step is an agent, and verification **rejects** a
proposal with no fidelity table, an empty one, one still carrying `<example>`, or
an over-cap rewrite with no stated reason.

**Demonstrated against the hand-done work** — clean-room against browser-tester
at origin/main, no sight of lane kp79's answers:

| | files | agent description chars | examples removed |
|---|---|---|---|
| before | — | 2,457 | 0 of 6 |
| hand sweep (kp79) | 3 | 1,664 | 6 of 6 |
| `refresh-descriptions` | **the same 3** | **1,644** | 6 of 6 |

Every pre-existing routing fact retained in both; totals within 1.2%. The one
real difference is named rather than averaged away.

**Creation surfaces** — `bundle-design-expert` taught three things that are now
wrong, including "Create awareness context — ~25-40 lines, domain exists,
delegate to expert", which is exactly the pointer-only file Phase 2.84 warns
about. Its own description was 739 chars, over the cap it teaches; rewritten to
591 with every routing fact kept. The skills bundle is held by another lane, so
its four patches ship as an artifact — and the defect is not hypothetical:
`personafy` currently *instructs* a description "at or below ~700-800
characters", nearly 2x the cap, and that bundle measures 13 of 31 skills over
400. Two clean-room arms differing only in the skill's own guidance:

| Arm | skillify | personafy | verdict |
|---|---|---|---|
| shipped | 453 | 714 | 2 WARNINGs |
| patched | 353 | 391 | **clean** |

## Recipe runs

| Target | Recipe | Result |
|---|---|---|
| foundation | validate-agents v1.8.0, **full** | ⚠ PASS WITH WARNINGS — 24 agents, 18 good / 5 polish / 1 needs_work / 0 critical; **0 errors** |
| anchors-amp-dev | validate-agents v1.8.0, **full** | ✅ PASS — 1 agent, 0 errors, 0 warnings |
| anchors-amp-dev | validate-bundle-repo v3.15.0, **full** (34 steps) | head cost 1,760 (lower bound); 0 new errors |
| foundation | validate-bundle-repo v3.15.0, deterministic steps | agents 0 err / 5 warn · skills 0 err / 1 warn · awareness 0 warn · head cost 3 warn, largest 26,368 |

Warnings are **quoted, not tuned**. The full `validate-bundle-repo` run writes
diagram files, so against the repo root only its deterministic steps were run —
named as a deviation in the DONE-NOTE, along with three findings reported and
not fixed (including one where a guard test caught me widening another lane's
exclusion contract, and I reverted).

## Tests

`pytest -q` → **2087 passed, 3 skipped, 2 failed**. The 2 failures are the two
known pre-existing ones, **confirmed** failing identically at origin/main
4384805 in a clean worktree.

Full working: `docs/lanes/pwmy-principles-to-tooling/DONE-NOTE.md`.

Refs: model_performance-pwmy

🤖 Draft — do not merge; the manager merges.

### DIFF
--- a/agents/bundle-design-expert.md
+++ b/agents/bundle-design-expert.md
@@ -2,11 +2,12 @@
 meta:
   name: bundle-design-expert
   description: |
-      **THE authoritative expert for designing, modeling, and BUILDING Amplifier bundles** — owns the full lifecycle from mechanism selection through behavioral modeling to YAML authoring and implementation.
+      Designing or building an Amplifier bundle: mechanism selection, bundle and behavior YAML, agent files, context architecture, behavioral modeling.
 
-      Use PROACTIVELY when: designing a new bundle, selecting mechanisms, writing bundle YAML or behaviors, authoring agent files (meta.description, WHY/WHEN/WHAT/HOW), making context architecture decisions (context sink, thin pointer, zero poisoning), or running behavioral modeling recipes.
+      USE WHEN a bundle, behavior, agent file or description is being authored or refreshed to the current rules, or a bundle needs modeling before implementation.
+      DO NOT USE WHEN the question is what foundation already ships (use foundation-expert), or is kernel and module internals (use core-expert).
 
-      **Authoritative on:** bundle design, mechanism selection, behavioral modeling, YAML authoring, behaviors, agent file authoring, context sink pattern, thin pointer, zero poisoning, bundle lifecycle, bundle anti-patterns, objectives-to-model recipes
+      **Authoritative on:** description authoring, awareness files, context sink, thin pointer, zero poisoning, bundle lifecycle, bundle anti-patterns
 model_role: general
 
 tools:
@@ -143,7 +144,7 @@ context:
 - If behavior is NOT composed → zero context about that capability
 - No partial knowledge, no context poisoning
 
-### The Thin Awareness Pointer
+### The Thin Awareness Pointer -- and when NOT to write one at all
 
 Root sessions should get just enough context to:
 1. Know a capability/domain exists
@@ -151,7 +152,23 @@ Root sessions should get just enough context to:
 3. NOT enough to attempt the work themselves
 
 **Anti-pattern:** 80 lines of "how bundles work" in always-loaded context
-**Correct pattern:** 25 lines saying "bundles exist, delegate to expert"
+
+**But first ask whether the file should exist at all.** The agent catalog and
+the visible-skills block ALREADY say a capability exists and when to use it,
+in a line that is paid anyway. An awareness file that only restates a catalog
+entry is the same sentence bought twice, and
+`validate-bundle-repo.yaml` Phase 2.84 now warns on exactly that
+(`awareness_redundant_with_catalog`, `awareness_is_pointer_only`).
+
+Write an awareness file ONLY for a **concept that has no catalog line of its
+own**: a cross-cutting hazard that belongs to no single capability, a routing
+table between this bundle and others, or a prerequisite the session needs
+before any of the capabilities can work. If what you are about to write is
+"agent X exists, delegate to it when Y" -- that is X's own
+`meta.description`. Put it there instead, and ship no awareness file.
+
+Full rule: @foundation:docs/BUNDLE_GUIDE.md, "Awareness: concept + trigger
++ pointer".
 
 ### Zero Context Poisoning
 
@@ -189,8 +206,12 @@ The system automatically searches the `/agents` directory relative to the bundle
 
 When building a new bundle with expert capabilities:
 
-1. **Create behavior YAML** -- includes agent + thin context pointer
-2. **Create awareness context** -- ~25-40 lines, domain exists, delegate to expert
+1. **Create behavior YAML** -- includes the agent; a context pointer only if
+   step 2 earns one
+2. **Create awareness context ONLY IF there is a concept the agent's own
+   `meta.description` cannot carry** (a cross-cutting hazard, a routing table,
+   a prerequisite). If there is not, ship no awareness file -- the catalog
+   line already does the job and is paid anyway
 3. **Create agent file** -- heavy @mentions to full documentation (context sink)
 4. **Ensure nothing in shared context** -- your domain knowledge is opt-in via composition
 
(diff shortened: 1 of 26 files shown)


bkrabach:
## Manager verification — all four scopes re-checked in a scratch clone against the **merge-base**. Merging.

Merge-base `4384805`, head `d55dacb`. **CI green on all 6 legs** (ubuntu + Windows × 3.11/3.12/3.13).

This is the PR that turns this batch's findings from convention into enforcement, so I checked each scope against its own artifact rather than the DONE-NOTE.

### (a) Docs — the rules, scoped to every bundle, with the numbers

`context/shared/description-authoring-principles.md:12` — **"Scope: EVERY bundle, not just this repository."** That single line is the fix for the defect this whole sweep exposed: #341's no-`<example>` policy existed only as foundation's local convention and **quietly stopped being true everywhere else**, which the file now says at `:35`.

The caps table at `:124-125` is stated with its enforcement point, not as advice:

| surface | WARN | ERROR (2×) | enforced by |
|---|---|---|---|
| Agent `meta.description` | > 600 | > 1,200 | `validate-agents.yaml`, `validate-bundle-repo.yaml` |
| Skill frontmatter `description` | > 400 | > 800 | `validate-bundle-repo.yaml` Phase 2.82 |

`docs/BUNDLE_GUIDE.md:378` carries **`## Awareness: concept + trigger + pointer`**; `docs/AGENT_AUTHORING.md:189` carries *"Awareness files are not a place to describe an agent"*. The WHY is measured, not asserted — `:27` records that **the runtime-appended agent catalog alone was 12,904 chars**, and the citations are all present (`84,319`, `48,249`, `29 files`, `6 repos`, `2ef5e12`, `ux32`, `#316`).

### (b) Validators — four checks, and the thresholds are argued from data

**Awareness redundancy (R1/R2)** is measurable and was *tested against reality before being trusted*: **R1 term-overlap ≥ 60%**, **R2 pointer-only** (the file invokes `delegate()`/`load_skill()` naming the capability). Run across foundation's own 36 context files, **R1 fires 0 times (max coverage observed 0.333) and R2 fires 3 times** — each on a file that genuinely is when-to-use prose plus a pointer. **`best_coverage` is reported for every file whether or not it fires, so the threshold can be re-argued against real numbers rather than re-guessed.**

**Head cost (Phase 2.86), WARNING at 4,000** — justified against the measured population, which is the part that makes it defensible:

```
engineered:    anchors 2,483   anchors-amp-dev 1,760
unengineered:  5,479   14,198 (foundation's own root)   15,002   26,368   29,684   87,555
```

**"4,000 is the smallest round number above every engineered bundle and below every unengineered one."** WARNING rather than ERROR **because foundation's own root bundle exceeds it — "that is a design conversation, not a build break."** An unresolvable include is **not charged a flat guess**: the total is marked `is_lower_bound` and the unresolved refs are named.

### (b5) Fail-before — reproducible by anyone, not quoted from a lane note

`test-fixtures/misaligned-bundle/` **violates all four checks by construction**, and its README says why it exists in exactly those terms. `tests/test_description_alignment_checks.py` (**24 tests**) pins it:

```
test_fixture_violates_all_four_checks()
    assert "example_block_present"            in types_of(agents,   "errors")
    assert "example_block_present"            in types_of(skills,   "errors")     ← the skill half, previously unenforced everywhere
    assert "awareness_redundant_with_catalog" in types_of(awareness,"warnings")
test_agent_description_over_2x_is_an_error / at_the_cap_warns / under_the_cap_is_clean
test_validate_agents_agrees_with_validate_bundle_repo_on_the_cap
```

That last one is the parity guard this program has needed twice before (`xe1u`/`dfni`): **the two recipes cannot drift to different caps.**

### (c) Creation skills — compliant by default

`skills/creating-amplifier-modules/SKILL.md:105` now emits the shape itself:

> `"What this tool does. USE WHEN <condition>. DO NOT USE WHEN <condition> -- use <alternative>."`

with a checklist item at `:302` asking whether the description leads with the trigger, names the alternative, and carries zero `<example>` blocks. `agents/bundle-design-expert.md` documents the rules at `:258-263` **and its own description now obeys them** — trigger-first with an explicit `DO NOT USE WHEN` naming `foundation-expert` for the rejected case. A guide that violated its own rule would have been the weakest possible version of this PR.

### (d) Refresh path

`recipes/refresh-descriptions.yaml` emits `proposals.md` — **a compliant rewrite for each violation WITH A FIDELITY TABLE** — and `verdict.json`, which re-checks the rewrites against the same thresholds **plus a structural gate on the fidelity tables themselves**. So an old bundle can be brought into alignment without a human having to trust an agent's summary of what it changed.

Merging under the owner's Stage-1 rule.

PR #376, opened 2026-09-07
perf(tool-delegate): strip <example>/<commentary> at catalog-render time -- -9,474 chars every request

bkrabach:
Closes `model_performance-6f80`.

## Why the renderer, and not more content edits

`context/shared/description-authoring-principles.md` V3 sets the policy: zero
`<example>` blocks, zero `<commentary>` tags, in any description surface. PR
#341 set it and it was applied inside amplifier-foundation and **nowhere
else**, because nothing at render time said otherwise. `kp79`/`kv98` then
stripped those blocks **by hand** from eight repos.

That fixes today's catalog and holds exactly until someone composes a bundle
this program does not own. **Measured on a real host right now: 25 `<example>`
blocks are still rendering into the delegate description**, every one from a
repo outside our reach. Content hygiene is necessary and not sufficient.

Every agent's `meta.description` is concatenated into this tool's own
description, which loads on **every turn of every session**, whether or not
that agent is ever delegated to.

**Strip at RENDER, not at load.** The description on disk stays whatever its
author wrote — we are not silently rewriting other people's files, we are
declining to pay for their examples on every request.

## Measured saving (owner's own app list, $0, no API call)

Both arms through the **same** renderer: the BEFORE arm neutralises
`_strip_example_blocks` to the identity function, which is byte-for-byte what
main does.

| | chars |
|---|---:|
| delegate description **before** | **49,954** |
| delegate description **after** | **40,480** |
| **saved** | **9,474 (18.97%)** |
| `<example` tags | 25 → 0 |
| `<commentary` tags | 13 → 0 |
| agents | 68 (10 stripped) |

```
2371  amplifier-online:deployment-guide           4004 -> 1633   6 examples
1380  app-cli:cli-expert                          2659 -> 1279   3
 969  notify:notify-expert                        1825 ->  856   2
 918  amplifier-tester:setup-digital-twin         1565 ->  647   2
 837  amplifier-tester:validator                  1342 ->  505   2
 717  digital-twin-universe:dtu-profile-builder   1468 ->  751   2
 605  infographic-builder:infographic-builder     1639 -> 1034   2
 600  terminal-tester:terminal-debugger           1270 ->  670   2
 565  terminal-tester:terminal-visual-tester      1222 ->  657   2
 512  terminal-tester:terminal-operator           1109 ->  597   2
```

Reproducer (re-runs at $0):
`docs/lanes/6f80-delegate-example-strip/evidence/measure_catalog_bytes.py`.

**Calibration:** the item's independent census of that host says 25 `<example>`
blocks. This reconstruction finds **exactly 25**. That is what makes the rest
of the table quotable.

**The item predicted "~20,000+" and the gap is arithmetic in the prediction,
not shortfall in the patch.** Its per-bundle table measures whole catalog
*entries*, not example payload — `terminal-tester 3,601+121+3 = 3,725`,
`app-cli 2,659+24+1 = 2,684`, `notify 1,825+26+1 = 1,852`,
`infographic 1,639+45+1 = 1,685`: four of five reconcile **to the character**.
The blocks are 9,474 of the 19,103 chars those ten entries occupy. Recorded in
the lane note rather than absorbed.

## FAIL-BEFORE — both counts

Same file, same command, merge-base `5f0f04b` vs branch:

```
at main     4 failed, 7 passed in 0.07s
on branch  11 passed in 0.05s
```

Verbatim from main's run — the fixture's example payload *in the rendered
catalog entry*:

```
AssertionError: assert 'How does auth work here?' not in '  - fixture...NE SURVIVES.'
  'How does auth work here?' is contained here:
    user: 'How does auth work here?'
    <commentary>
```

## Fidelity is the whole risk, so it is tested five ways

A regex that eats a paragraph past the closing tag is the failure mode.
`TestNothingOutsideTheBlocksIsConsumed` pins the boundary:

1. Every line of surrounding prose survives — trigger, USE WHEN, DO NOT USE
   WHEN, "Authoritative on", and a final line after the last block.
2. `BEFORE <example>eaten</example> AFTER` keeps `BEFORE ` and ` AFTER`
   byte-for-byte. `.*?` is non-greedy; the optional trailing `[ \t]*\n` fires
   only when it ends the line.
3. A description with **no** blocks renders byte-identical — guarded twice
   (regex, plus an early return).
4. An **unpaired** `<example>` matches nothing and is left alone. Swallowing
   the rest of a description because its author forgot a close tag would delete
   exactly the routing facts the catalog exists to carry.
5. Other agents in the same catalog are untouched.

Corpus-wide over all 68 discovered agents: **0 surviving lines missing from the
render**.

## The word is not the tag (amendment)

The five boundary tests above all use fixtures that carry real
`<example>`/`<commentary>` tags. None of them contained the word "example"
*outside* a tag — so none of them could tell a tag-keyed strip from a
word-keyed one. `test_a_description_with_no_blocks_is_byte_identical` is the
closest, and its fixture contains neither the tag nor the word.

That left the corpus's most common shape untested: ordinary prose that says
"example" and "commentary" out loud and has no tags at all. Added, as a named
fixture plus three assertions, in the same file:

```
tests/test_delegate_example_strip.py::TestOrdinaryProseSayingExampleIsUntouched::test_ordinary_prose_renders_byte_identical PASSED [ 33%]
tests/test_delegate_example_strip.py::TestOrdinaryProseSayingExampleIsUntouched::test_every_line_of_the_ordinary_prose_survives PASSED [ 66%]
tests/test_delegate_example_strip.py::TestOrdinaryProseSayingExampleIsUntouched::test_ordinary_prose_is_not_reported_as_stripped PASSED [100%]

3 passed in 0.08s
```

The rendered entry must be **byte-identical** — same bytes, same newlines — and
the debug line must not name an agent nothing was stripped from.

**This is a regression pin, not a fail-before, and that is stated plainly**: it
is green at main too, since main strips nothing. Its worth was measured by
mutating `_strip_example_blocks` instead:

| mutant | result |
|---|---|
| guard word-keyed, regex still tag-keyed | 14/14 still pass — the widened guard alone is unobservable |
| regex word-keyed, guard still tag-keyed | these 3 still pass — the tag-keyed guard short-circuits first |
| **both word-keyed** | **these 3 fail.** Two older tests fail too, but only on their tagged fixtures; the ordinary prose is deleted silently and this class is the only thing that sees it |

The middle row is this pin's honest limit, not omitted: it pins the
guard-and-regex contract as a pair, not the regex alone.

Nothing existing was changed, weakened, deleted or moved — the diff is purely
additive apart from renumbering the observability section 3 → 4, and the file
stays in `tests/`, the path CI runs.

## The strip names who it stripped

```
DEBUG amplifier_module_tool_delegate: delegate catalog: stripped
<example>/<commentary> blocks from 10 agent description(s) at render time
(the descriptions on disk are unchanged): amplifier-online:deployment-guide,
amplifier-tester:setup-digital-twin, amplifier-tester:validator,
app-cli:cli-expert, digital-twin-universe:dtu-profile-builder,
infographic-builder:infographic-builder, notify:notify-expert,
terminal-tester:terminal-debugger, terminal-tester:terminal-operator,
terminal-tester:terminal-visual-tester
```

Names, not a count. A bundle author whose examples vanish can find out why —
otherwise this trades one silent behaviour for another.
`test_no_log_line_when_nothing_was_stripped` pins that a no-op strip is silent.

## Suite

| | main (`5f0f04b`, patch stashed) | branch |
|---|---|---|
| `uv run pytest -q` | 4 failed, 2095 passed, 4 skipped | **2102 passed, 4 skipped, 0 failed** |
| CI's own command (`pytest tests/ -q`) | — | **1912 passed, 4 skipped** |
| `tests/test_delegate_example_strip.py` | — | **14 passed** (was 11) |
| `ruff check` (changed files) | — | All checks passed |
| `ruff format --check` (changed files) | — | already formatted |

Main's four failures are precisely this PR's four fail-before assertions.
Branch failure set is empty. The `2099 -> 2102` is the amendment's
three tests and nothing else.

**CI 6/6 green on the head**, verbatim from the fetched log of run
[34164562161](https://github.com/microsoft/amplifier-foundation/actions/runs/34164562161):

```
Tests (ubuntu-latest,  Python 3.11)  success   1912 passed, 4 skipped, 1 warning in 26.23s
Tests (ubuntu-latest,  Python 3.12)  success   1912 passed, 4 skipped, 1 warning in 39.80s
Tests (ubuntu-latest,  Python 3.13)  success   1912 passed, 4 skipped, 1 warning in 37.80s
Tests (windows-latest, Python 3.11)  success   1907 passed, 9 skipped, 1 warning in 55.72s
Tests (windows-latest, Python 3.12)  success   1907 passed, 9 skipped in 65.59s
Tests (windows-latest, Python 3.13)  success   1907 passed, 9 skipped, 1 warning in 55.40s
```

Counts identical to the run before `-rA` landed — the flag changed what the log
says, not what CI runs.

The one warning is pre-existing and unrelated (`AsyncMockMixin` coroutine
warning from `tests/test_subprocess_runner.py`, present in the previous run
too). The amendment's own file runs clean: `14 passed in 0.04s`.

## One finding for the reviewer — CI cannot see the module's own tests

`pyproject.toml` sets `testpaths = ["tests", "modules/tool-delegate/tests"]`,
but `.github/workflows/ci.yml` runs `uv run pytest tests/ -q --tb=short`. An
explicit path argument **overrides `testpaths` entirely**, so all 17 test files
under `modules/tool-delegate/tests/` are invisible to every one of the six CI
legs today.

This item exists because a rule with no enforcement point decays. A fail-before
test CI never executes is that same failure wearing a test's costume — so this
test lives in `tests/`, following the precedent of
`tests/test_named_delegate_matrix_67u.py`.

**The CI gap itself is not fixed here** (out of scope, and it would change what
the whole suite covers). Fix is one of: add `modules/tool-delegate/tests` to
the CI command, or drop the path argument.

**Second consequence, found while amending — and fixed here.** `-q` printed a
dot per test and never a node id, so no green CI log could quote any test by
name:

```
$ grep -c "TestOrdinaryProseSayingExampleIsUntouched" <fetched log of run 34163908081>
0
```

I first called that unobtainable and published a count identity instead
(ubuntu `1909 -> 1912`, windows `1904 -> 1907`, +3 on every leg against an
amendment adding exactly three tests). That reasoning holds, but it is
arithmetic about a number standing in for evidence about a test — and `-q`
hiding node ids is a property of one flag, not a law of CI.

So `ci.yml` now runs `pytest tests/ -q --tb=short -rA`. `-rA` appends the
short-summary section naming every test by node id. **Nothing about which tests
run changes** — the explicit `tests/` path still overrides `testpaths`, the
module's 17 files are still invisible, and all six legs still report the same
1912/1907. Only the log gains ~1,916 named lines per leg.

Verbatim from the green log of run
[34164562161](https://github.com/microsoft/amplifier-foundation/actions/runs/34164562161)
(`a168973`), ubuntu 3.13 leg:

```
PASSED tests/test_delegate_example_strip.py::TestOrdinaryProseSayingExampleIsUntouched::test_ordinary_prose_renders_byte_identical
PASSED tests/test_delegate_example_strip.py::TestOrdinaryProseSayingExampleIsUntouched::test_every_line_of_the_ordinary_prose_survives
PASSED tests/test_delegate_example_strip.py::TestOrdinaryProseSayingExampleIsUntouched::test_ordinary_prose_is_not_reported_as_stripped
```

```
$ grep -c "TestOrdinaryProseSayingExampleIsUntouched" <fetched log of run 34164562161>
18            # 3 tests x 6 legs -- named on every leg, POSIX and Windows
```

**The `testpaths` gap above is still not fixed** — that is a separate change to
what CI executes, and it stays out of scope.

## Spend

**$0.00 of a $0.00 authority.** No API call, no DTU. The measurement is an
offline render over the host's existing read-only bundle cache; `~/.amplifier/`
was never written (`zc6t` F6: `amplifier source add` ignores `AMPLIFIER_HOME`).

Draft until CI is green on all six legs. **Do not merge from the lane — the
manager merges.**

Generated with Amplifier


bkrabach:
## Manager verification — render defense behaves correctly, but the required word-only regression test is missing. Returning to DRAFT.

I independently reproduced the functional gates in isolated `main`/branch checkouts:

* **Fail-before:** copied the branch fixture test to clean main: `4 failed, 7 passed`; clean branch: `11 passed`.
  Main rendered both `<example>` and `How does auth work here?`; branch rendered neither.
* **No mutation / fidelity:** branch’s test suite verifies surrounding bytes; the cache-catalog remeasurement is 68 agents, `49,954 → 40,480` chars (**−9,474 / 18.97%**), tags `25/13 → 0/0`, and 0 surviving lines missing.
* **Observability:** `delegate catalog: stripped <example>/<commentary> blocks from 10 agent description(s) at render time ...` names every affected agent.
* **Suite:** my branch run: `2099 passed, 4 skipped`; ruff check and format check pass. PR CI run 34160123860 is 6/6 green.

But gate 2 is specifically: *an agent description that merely contains the word `example` must be untouched, with a negative test.* The submitted `test_a_description_with_no_blocks_is_byte_identical` fixture is `"Design, architecture, planning, and code review..."`; it contains no ordinary `example` text. `test_other_agents_are_untouched` is `"Plain description, no blocks."`, also without it. I manually probed a word-only description and the implementation preserves it byte-identically, but that is not a durable regression test.

Add a named test fixture such as `"Use the example selector..."` with no `<example` tag, assert byte identity, rerun the suite/CI, then re-open. No scope expansion is needed.

PR #388, opened 2026-09-15
chore: drop the dead max_tokens from every bundle; restore minimal-delegate's 64k

bkrabach:
> **Merge this AFTER `context-simple`, `provider-gemini` and `app-cli`.** These deletions are only safe once context-simple's new semantics are in.

### 1. `chore: drop the dead max_tokens from every bundle`

context-simple's `max_tokens` used to be a fallback consulted only when a provider published no context window. Orchestrators always pass a provider, so these values never reached the wire — they were inert in every bundle here.

It is now a CAP defaulting to `None` (no cap). Leaving these lines in place would silently clamp every session to 300,000 tokens — including sessions on 1M-context models — turning a documentation fix into a 3x behavior change nobody asked for. Deleting them preserves today's behavior exactly, and leaves no fossil that looks meaningful.

Files, all `session.context.config.max_tokens`:

| File | Value removed |
|---|---|
| `bundle.md` | 300000 |
| `bundles/anchors/bundle.md` | 300000 |
| `experiments/delegation-only/bundle.md` | 300000 |
| `experiments/exp-foundation.md` | 300000 |
| `experiments/exp-lean/behaviors/lean-foundation.yaml` | 300000 |
| `experiments/minimal-delegate/.../minimal-delegate-foundation.yaml` | 64000 |

`anchors-amp-dev`, `amplifier-dev`, `minimal`, `with-anthropic` and `with-openai` set no `max_tokens` of their own — they inherit, so the root `bundle.md` covers them.

### 2. `feat(minimal-delegate): restore the 64k budget, now that it actually applies`

That experiment asked for `max_tokens: 64000` — a deliberately tight budget, the whole point of a delegate-only head — and never got it. The previous commit deleted the line to preserve behavior; restoring it is a **deliberate choice, not an inherited one**: the cap works, nobody depends on this bundle today, and the 64k constraint is finally measurable.

## How to verify

Every touched file was re-parsed after the edit to prove the YAML still loads and only the intended key is gone.

```
1938 passed, 3 skipped
```

1 pre-existing failure, **verified failing on `main` before this change**: `test_grpc_adapter_main.py::TestVerifyModuleType::test_non_isinstance_object_with_mount_passes`.

## Breaking changes

Behavior is preserved exactly for every bundle except `minimal-delegate`, which now genuinely runs under a 64,000-token cap — an intentional, opted-in change to an experiment with no current dependents.

## Coordinated five-repo change

`max_tokens` in `context-simple` was documented as "Maximum context size" but implemented as a fallback consulted only when a provider published no context window. Orchestrators always pass a provider, so the knob was silently dead in production. It is now a real **cap** (default `None` = no cap); a new `max_tokens_fallback` (default 200,000) took over the fallback role.

### Merge order matters

**Merge FIRST** (independent of each other, any order):
1. `amplifier-module-context-simple` — `feat/max-tokens-cap`
2. `amplifier-module-provider-gemini` — `feat/publish-context-window`
3. `amplifier-app-cli` — `feat/session-module-config-overrides`

**Merge AFTER those three:**
4. `amplifier-foundation` — `chore/drop-dead-max-tokens`
5. `amplifier-bundle-attractor` — `chore/drop-dead-max-tokens`

Rationale: the two config repos delete `max_tokens` lines that only become safe-to-delete once context-simple's new semantics are in.

## Cross-repo verification

DTU instance `context-overflow-fix-20260915`, all 8 target behaviors PASS:

- A Gemini session's effective budget goes from **200,000** (the old fallback) to **1,011,712** (its real published window) — a **5.1x** increase.
- `max_tokens: 500000` caps to exactly 500,000.
- A cap above the model window is a no-op.
- `settings.yaml` `overrides.context-simple.config` now reaches `session.context` through the CLI's own `resolve_bundle_config`.

| Repo | Result |
|---|---|
| context-simple | 166 passed, 1 xfailed |
| provider-gemini | 330 passed; 1 pre-existing failure (`test_image_vision_integration_with_real_api`, needs a live `GOOGLE_API_KEY`, fails on `main` too) |
| app-cli | 2335 passed, 2 skipped, 1 xfailed |
| amplifier-foundation | 1938 passed, 3 skipped; 1 pre-existing failure (`test_grpc_adapter_main.py::TestVerifyModuleType::test_non_isinstance_object_with_mount_passes`, verified failing on `main` before the change) |
| amplifier-bundle-attractor | 268 passed, 2 skipped |


bkrabach:
## The five PRs in this coordinated change

**Merge FIRST — independent of each other, any order:**
1. context-simple — https://github.com/microsoft/amplifier-module-context-simple/pull/42
2. provider-gemini — https://github.com/microsoft/amplifier-module-provider-gemini/pull/48
3. app-cli — https://github.com/microsoft/amplifier-app-cli/pull/342

**Merge AFTER those three:**
4. amplifier-foundation — https://github.com/microsoft/amplifier-foundation/pull/388
5. amplifier-bundle-attractor — https://github.com/microsoft/amplifier-bundle-attractor/pull/359

The two config repos (4, 5) delete `max_tokens` lines that only become safe-to-delete once context-simple's new cap semantics (1) are in.

